PR #30, opened 2020-09-28
rwa-53-unclaim-a-task

lewis150193:
https://tools.hmcts.net/jira/browse/RWA-53

### Change description ###
Unclaim a task

```
[ ] Yes
[ x ] No
```


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaService.java
```diff
@@ -57,4 +57,23 @@ public void claimTask(String taskId, String subjectId) {
         }
 
     }
+
+    public String unclaimTask(String id) {
+        try {
+            return camundaServiceApi.unclaimTask(id);
+        } catch (FeignException ex) {
+            if (HttpStatus.NOT_FOUND.value() == ex.status()) {
```
chrisgrimble, line 65: I think you can catch an exception of type FeignException.NotFound rather than checking this

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/TaskControllerTest.java
```diff
@@ -119,7 +129,7 @@ public void should_return_503_for_work_in_progress_endpoints() {
             .body("status", equalTo(HttpStatus.SERVICE_UNAVAILABLE.value()))
             .body("message", equalTo(responseMessage))
             .when()
-            .post("/task/{task-id}/unclaim", taskId);
+            .post("/task");
```
chrisgrimble, line 132: Not too sure why this has changed to /task you have been implementing /task/{task-id}/unclaim so why change this to /task?

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/TaskControllerTest.java
```diff
@@ -59,6 +59,16 @@ void should_succeed_and_return_a_204_no_content() {
...
     }
 
+    @Test
+    void should_unclaim_a_task_204_no_content() {
+
+        String taskId = UUID.randomUUID().toString();
+
+        ResponseEntity<String> response = taskController.unclaimTask(taskId);
+
+        assertNotNull(response);
```
chrisgrimble, line 69: Don't think we need to check not null.

chrisgrimble: Also do we not want to check we called something on the camundaService or something like that?

chrisgrimble (changes requested):
I assume all the commits on this PR by Alex are from branching from another PR so have just looked at changes by @lewis150193

PR #47, opened 2020-10-16
Updated Task

lewis150193:

https://tools.hmcts.net/jira/browse/RWA-158

### Change description ###

```
[ ] Yes
[ x ] No
```


iguannago (approved):

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/utils/CreateHmctsTaskVariable.java
```diff
@@ -0,0 +1,76 @@
...
+            caseId,
+            caseCategory,
+            caseName
+
+        );
+    }
+
+    private String getValueFromObject(Map<String, CamundaVariable> localVariableResponse, String key) {
+        return localVariableResponse.entrySet().stream()
+            .filter(e -> key.equals(e.getKey()))
```
chrisgrimble, line 71: Isn't this just localVariableResponse.get(key)

src/main/resources/application.yaml
```diff
@@ -62,6 +62,7 @@ security:
     - "/favicon.ico"
   authorisedServices:
     - "wa_task_management_api"
+    - "ccd_gw"
```
chrisgrimble, line 65: I'd expect we would need to allow ccd_something or other eventually but what made us make this change now? Is it part of the tests. Just interested but wondered if this got added and isn't really needed?

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/PojoTest.java
```diff
@@ -28,11 +28,12 @@
         Token.class,
         UserInfo.class,
         HistoryVariableInstance.class,
-        AddLocalVariableRequest.class
+        AddLocalVariableRequest.class,
```
chrisgrimble, line 31: I didn't know Java allowed you to do that will stop this class always conflicting if two people add classes to it at the same time

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/domain/entities/camunda/MappedTaskTest.java
```diff
@@ -0,0 +1,61 @@
...
+        Assertions.assertThat(mappedTask.getLocation()).isEqualTo("some-location");
+        Assertions.assertThat(mappedTask.getLocationName()).isEqualTo("some-location-name");
+        Assertions.assertThat(mappedTask.getExecutionType()).isEqualTo("some-executionType");
+        Assertions.assertThat(mappedTask.getJurisdiction()).isEqualTo("some-jurisdiction");
+        Assertions.assertThat(mappedTask.getRegion()).isEqualTo("some-region");
+        Assertions.assertThat(mappedTask.getCaseTypeId()).isEqualTo("some-caseTypeId");
+        Assertions.assertThat(mappedTask.getCaseId()).isEqualTo("some-caseId");
+        Assertions.assertThat(mappedTask.getCaseCategory()).isEqualTo("some-cat");
+        Assertions.assertThat(mappedTask.getCaseName()).isEqualTo("some-case");
+        Assertions.assertThat(mappedTask.getAutoAssigned()).isTrue();
```
chrisgrimble, line 58: No need for this if we have added it to the PojoTest

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/domain/enums/AssigneeEnumTest.java
```diff
@@ -0,0 +1,28 @@
+package uk.gov.hmcts.reform.wataskmanagementapi.domain.enums;
+
+import org.junit.jupiter.api.Test;
+import uk.gov.hmcts.reform.wataskmanagementapi.domain.entities.camunda.TaskState;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class AssigneeEnumTest {
+
```
chrisgrimble, line 9: No need for this

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaServiceTest.java
```diff
@@ -52,37 +53,48 @@
...
+        camundaService.getTask(taskId);
 
-        CamundaTask response = camundaService.getTask(taskId);
+        verify(camundaServiceApi, times(1)).getTask(eq(taskId));
+        verify(camundaServiceApi, times(1)).getVariables(camundaTask.getId());
 
-        verify(camundaServiceApi, times(1)).getTask(taskId);
         verifyNoMoreInteractions(camundaServiceApi);
 
-        assertEquals(mockedTask, response);
```
chrisgrimble, line 85: We should be mocking the call the camundaServiceApi, then the call to createHmctsTaskVariable.mapToTaskObject and checking the response is what is returned by this method.

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskMapperTest.java
```diff
@@ -35,7 +35,8 @@ void should_allow_null() {
             ZonedDateTime.now(),
```
chrisgrimble, line 35: Not sure the TaskMapper is needed anymore

src/testUtils/java/uk/gov/hmcts/reform/wataskmanagementapi/config/GivensBuilder.java
```diff
@@ -102,6 +102,20 @@ public GivensBuilder iAddVariablesToTaskWithId(String taskId, CamundaProcessVari
         return this;
     }
 
+    public GivensBuilder iAddLocalVariablesToTaskWithId(String taskId, CamundaProcessVariables processVariables) {
+        given()
+            .contentType(APPLICATION_JSON_VALUE)
+            .baseUri(camundaUrl)
+            .body(new Modifications(processVariables.getProcessVariablesMap()))
+            .when()
+            .post("/task/{task-id}/variables", taskId)
```
chrisgrimble, line 111: I think you should use /task/{task-id}/localVariables. I think the documentation for /task/{task-id}/variables is a bit vague when dealing with variables that do not already exist. As we know these should be task local variables better to use the /localVariables endpoint. Then we can be sure we are setting a task local variable not a process variable.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/domain/entities/camunda/MappedTask.java
```diff
@@ -0,0 +1,175 @@
...
+@ApiModel("Task")
+public class MappedTask {
+
+    private String id;
+    private String name;
+    private String assignee;
+    private String type;
+    private String taskState;
+    private String taskSystem;
+    private String securityClassification;
```
chrisgrimble, line 24: This can be an enum

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/domain/entities/camunda/MappedTask.java
```diff
@@ -0,0 +1,175 @@
...
+    private String executionType;
+    private String jurisdiction;
+    private String region;
+    private String caseTypeId;
+    private String caseId;
+    private String caseCategory;
+    private String caseName;
+    private Boolean autoAssigned;
+
+    public MappedTask() {
```
chrisgrimble, line 43: We have said we will document this interface for CCD well. You can do that by adding annotations to this class which get rendered on the swagger page. See CreateTaskRequest in wa-workflow-api and the @ApiModelProperty annotation. We need to describe each field and also say any restrictions on the value i.e. give the format for the strings returned by dueDate etc.

chrisgrimble (changes requested):
Added some comments

iguannago (approved):

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/clients/CamundaServiceApi.java
```diff
@@ -90,4 +90,12 @@ void assigneeTask(@PathVariable("task-id") String id,
     @ResponseBody
     Map<String, CamundaVariable> getVariables(@PathVariable("task-id") String id);
 
+    @GetMapping(
+        value = "/task/{task-id}/localVariables",
+        produces = APPLICATION_JSON_VALUE
+    )
+    @ResponseBody
+    Map<String, CamundaVariable> getLocalVariables(@PathVariable("task-id") String id);
```
alexbaeza, line 98: I don't think we need this endpoint for the task mapper instead we could use:
`/task/{task-id}/variables`
which should return all variables, the reason for this is performance, when we map a task we need everything in camunda to map it to our Task object, in this case: we could have some variables that could only be retrieved through the /variables endpoint so we will have to query both endpoints `/localVariables` and `/variables`
so might be better to use `/variables` as it should contain all variables, not just the local ones.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/domain/entities/camunda/TaskState.java
```diff
@@ -0,0 +1,22 @@
...
+
+    private final String taskState;
+
+    TaskState(String taskState) {
+        this.taskState = taskState;
+    }
+
+    public String  getTaskState() {
+        return this.taskState;
+    }
```
alexbaeza, line 21: should this be toString()

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/utils/TaskMapper.java
```diff
@@ -0,0 +1,80 @@
+package uk.gov.hmcts.reform.wataskmanagementapi.utils;
```
alexbaeza, line 1: We need to restore the previous deleted class and place this one in:
package uk.gov.hmcts.reform.wataskmanagementapi.services;
Since this is a service, not util class :)

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/utils/TaskMapper.java
```diff
@@ -0,0 +1,80 @@
...
+import uk.gov.hmcts.reform.wataskmanagementapi.domain.entities.camunda.CamundaObjectMapper;
+import uk.gov.hmcts.reform.wataskmanagementapi.domain.entities.camunda.CamundaTask;
+import uk.gov.hmcts.reform.wataskmanagementapi.domain.entities.camunda.CamundaVariable;
+import uk.gov.hmcts.reform.wataskmanagementapi.domain.entities.camunda.Task;
+
+import java.util.Map;
+import java.util.Optional;
+
+
+@SuppressWarnings("PMD.UnnecessaryConstructor")
```
alexbaeza, line 14: I don't think this suppression is needed

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/utils/TaskMapper.java
```diff
@@ -0,0 +1,80 @@
...
+        this.camundaObjectMapper = camundaObjectMapper;
+    }
+
+    public Task mapToTaskObject(Map<String, CamundaVariable> localVariableResponse, CamundaTask camundaTask) {
+        // Camunda Attributes
+        String id = camundaTask.getId();
+        String name = camundaTask.getName();
+        String task = camundaTask.getFormKey();
+        String createdDate = camundaTask.getCreated().toString();
+        String dueDate = camundaTask.getDue().toString();
```
alexbaeza, line 32: If for some reason getCreated() or getDue() return null you'll have a `NullPointerException` since you can't call toString()

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/utils/TaskMapper.java
```diff
@@ -0,0 +1,80 @@
...
+        String assignee = camundaTask.getAssignee();
+        // Local Variables
+        String taskState = getValueFromObject(localVariableResponse.get("taskState"), String.class);
+        String securityClassification = getValueFromObject(localVariableResponse
+                                                               .get("securityClassification"), String.class);
+        String taskTitle = getValueFromObject(localVariableResponse.get("title"), String.class);
+        String executionType = getValueFromObject(localVariableResponse.get("executionType"), String.class);
+        boolean autoAssigned = false;
+        String taskSystem = getValueFromObject(localVariableResponse.get("taskSystem"), String.class);
+        String jurisdiction = getValueFromObject(localVariableResponse.get("jurisdiction"), String.class);
```
alexbaeza, line 42: Following on from the earlier comment about calling the /variables endpoint could we also rename these to what it was:

`getVariableValue(variables.get("ccdId"), String.class);`

src/main/resources/application.yaml
```diff
@@ -71,6 +71,6 @@ feign:
   client:
     config:
       default:
-        connectTimeout: 30000
-        readTimeout: 30000
+        connectTimeout: 50000
+        readTimeout: 50000
```
alexbaeza, line 75: we will deal with this in another ticket but let's keep the time out to 30 secs

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/domain/entities/CamundaTaskTest.java
```diff
@@ -27,7 +27,8 @@ void should_create_full_object_and_get_values() {
             created,
             dueDate,
             "some-description",
-            "some-owner"
+            "some-owner",
+            "formKey"
         );
```
alexbaeza, line 33: I believe we need more assertions here :D as there are a lot more properties in our Task object

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/domain/enums/AssigneeEnumTest.java
```diff
@@ -0,0 +1,28 @@
...
+    @Test
```
alexbaeza, line 10: I mean having a test for this isn't bad but perhaps let's also add another test to assert the size of TaskState that way if someone adds something he'll need to remember to update the test.

also TaskState.ASSIGNED.getTaskState() should be toString()

src/testUtils/java/uk/gov/hmcts/reform/wataskmanagementapi/config/GivensBuilder.java
```diff
@@ -104,6 +104,20 @@ public GivensBuilder iAddVariablesToTaskWithId(String taskId, CamundaProcessVari
...
 
+    public GivensBuilder iAddLocalVariablesToTaskWithId(String taskId, CamundaProcessVariables processVariables) {
+        given()
+            .contentType(APPLICATION_JSON_VALUE)
+            .baseUri(camundaUrl)
+            .body(new Modifications(processVariables.getProcessVariablesMap()))
+            .when()
+            .post("/task/{task-id}/localVariables", taskId)
+            .then()
+            .log().all(true)
```
alexbaeza, line 115: no need to log

alexbaeza (changes requested):

alexbaeza (commented):
LGTM 👀

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskSearchControllerTest.java
```diff
@@ -94,9 +94,9 @@ public void should_return_a_200_with_search_results_based_on_jurisdiction_and_lo
 
         result.then().assertThat()
             .statusCode(HttpStatus.OK.value())
-            .body("tasks.state", everyItem(either(is("unassigned")).or(is("assigned"))))
-            .body("tasks.case_data.reference", hasItem(task.get("ccdId")))
-            .body("tasks.case_data.location.id", everyItem(equalTo("17595")));
+            .body("tasks.task_state", everyItem(either(is("unassigned")).or(is("assigned"))))
+            .body("tasks.case_id", hasItem(task.get("ccdId")))
+            .body("tasks.location", everyItem(equalTo("17595")));
     }
```
BenNzewi, line 101: do we need a functional test to cover the get Acceptance criteria? i.e. GET /task/{task-id} 
To also show/assert all the standard attributes

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskSearchControllerTest.java
```diff
@@ -65,8 +64,9 @@ public void should_return_a_200_with_search_results() {
 
         result.then().assertThat()
             .statusCode(HttpStatus.OK.value())
-            .body("tasks.case_data.jurisdiction", everyItem(is("IA")))
-            .body("tasks.case_data.reference", hasItems(task1.get("ccdId"), task2.get("ccdId")));
+            .body("tasks.jurisdiction", everyItem(is("IA")))
+            .body("tasks.case_id", hasItem(task.get("ccdId")))
+            .body("tasks.id", hasItem(task.get("taskId")));
```
BenNzewi, line 69: Can we include all standard attributes in at least one of these put calls to cover the AC for POST /task(search with info)?

BenNzewi (dismissed):

BenNzewi (approved):

alexbaeza (approved):

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/domain/entities/camunda/Task.java
```diff
@@ -0,0 +1,249 @@
...
+import io.swagger.annotations.ApiModelProperty;
+import org.springframework.format.annotation.DateTimeFormat;
+
+import java.time.ZonedDateTime;
+import java.util.Objects;
+
+import static uk.gov.hmcts.reform.wataskmanagementapi.domain.entities.camunda.CamundaTime.CAMUNDA_DATA_TIME_FORMAT;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+@SuppressWarnings({"PMD.LawOfDemeter","PMD.TooManyFields","PMD.ExcessiveParameterList","PMD.ShortClassName"})
```
iguannago, line 15: Let's try to address the PMD. Otherwise, what's the point to use it
There are actually too many fields, and the constructor has too many parameters, right? Probably, it's useful to use a builder or to create another POJOs to organise similar semantic fields.

alexbaeza (approved):

PR #51, opened 2020-10-22
RWA-182: Search API refinement

alexbaeza:

https://tools.hmcts.net/jira/browse/RWA-182

### Change description ###

Fixes an issue with Jackson overriding camelCase to snake case even when using different configuration within the feign client, fixed by defining a decoder and encoder ObjectMapper and adding it as configuration to the feign client, this was causing issues when retrieving search results from camunda eventually returning everything as it was ignoring the other properties.

Added new structure to the search query and added an operator property that would be used in the future to define the explicitly operations and allows for better future improvements.

Removing ccdId eventId pre event state and post event state from the queries as these will be covered by another ticket.

Minor fixes to default constructors.

Added UnsupportedOperationException in callback controller to map to a 400, this is useful when using operators that we do not  yet support

```
[ ] Yes
[ x] No
```


lewis150193 (approved):

iguannago (approved):

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/TaskControllerTest.java
```diff
@@ -193,31 +198,25 @@ public void should_return_a_200_with_search_results() {
...
             .post("task");
 
         result.then().assertThat()
             .statusCode(HttpStatus.OK.value())
-            .body("tasks.size()", equalTo(1))
-            .body("tasks[0].caseData.reference", equalTo(ccdId2))
+            .body("tasks.caseData.jurisdiction", everyItem(is("IA")))
+            .body("tasks.caseData.reference", contains(ccdId2, ccdId1))
         ;
     }
```
BenNzewi, line 222: Can we add a test that cover OR operand. i.e. AC 3, e.g. (Then the Search API treats each parameter as if they are joined by a boolean OR (e.g. Location="12344567" OR Location="5356363")

alexbaeza: Added extra tests to cover that scenario

BenNzewi (dismissed):

iguannago (approved):

PR #52, opened 2020-10-23
Fix assignee endpoint

lewis150193:
### Change description ###
Fix assignee endpoint from assignee to assign and test fix

```
[ ] Yes
[ x ] No
```


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaService.java
```diff
@@ -82,14 +82,10 @@ public void assigneeTask(String taskId, String userId) {
...
-                    taskId
-                ), ex);
-            } else {
-                camundaErrorDecoder.decodeException(ex);
-            }
+            throw new ResourceNotFoundException(String.format(
+                "There was a problem assigning the task with id: %s",
+                taskId
+            ), ex);
         }
```
alexbaeza, line 89: I believe you're making two calls here:
`camundaServiceApi.addLocalVariablesToTask(taskId, camundaLocalVariables);`
and
`camundaServiceApi.assigneeTask(taskId, body);`

the first one could potentially return a 200 or a 500 according to camunda 500 is either the task id is null or doesnt exist in which case we should map to a 404.

however for the second call: 

`camundaServiceApi.assigneeTask(taskId, body);`

Could return a 204 is successful or a 500 meaning the task id was not found or something went wrong when assigning the task. Since we verified the task existed on the first call if a 500 is thrown it is safe to assume that something went wrong when assigning the task and therefore we should return a 500.
So perhaps we need to handle these 2 calls and it's exceptions separately

alexbaeza (changes requested):

alexbaeza (approved):

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaServiceTest.java
```diff
@@ -338,12 +339,51 @@ void assigneeTask_should_throw_resource_not_found_exception_when_other_exception
...
 
-        assertThatThrownBy(() -> camundaService.assigneeTask(taskId, userId))
+        assertThatThrownBy(() -> camundaService.assignTask(taskId, userId))
+            .isInstanceOf(ServerErrorException.class)
+            .hasCauseInstanceOf(FeignException.class);
+
+    }
+
+    @Test
+    void assigneeTask_should_throw_decoded_exception_when_other_exception_is_thrown() {
```
BenNzewi, line 351: should this be **assign**Task_should_throw_decoded_exception_when_other_exception_is_thrown and not **assignee**Task_should_throw_decoded_exception_when_other_exception_is_thrown?

BenNzewi (approved):

PR #83, opened 2020-11-20
search for task completion

lewis150193:
https://tools.hmcts.net/jira/browse/RWA-111

### Change description ###
Add Endpoint for search task using DMN.

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskForSearchCompletionControllerTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskForSearchCompletionControllerTest.java
@@ -1,0 +1,124 @@
+package uk.gov.hmcts.reform.wataskmanagementapi.controllers;
+
+import io.restassured.response.Response;
+import org.junit.Test;
+import org.springframework.http.HttpStatus;
+import uk.gov.hmcts.reform.wataskmanagementapi.SpringBootFunctionalBaseTest;
+import uk.gov.hmcts.reform.wataskmanagementapi.domain.entities.camunda.CamundaVariableDefinition;
+import uk.gov.hmcts.reform.wataskmanagementapi.domain.entities.idam.SearchEventAndCase;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Map;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.lessThanOrEqualTo;
+import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
+
+public class PostTaskForSearchCompletionControllerTest extends SpringBootFunctionalBaseTest {
+
+    private static final String ENDPOINT_BEING_TESTED = "task/completableByCaseEvent";
+
+    @Test
+    public void should_return_a_403_when_the_user_did_not_have_any_roles() {
+        Map<String, String> task = common.setupTaskAndRetrieveIds();
+
+        SearchEventAndCase searchEventAndCase = new SearchEventAndCase(task.get("caseId"), "requestRespondentEvidence");
+
+        Response result = restApiActions.post(
+            ENDPOINT_BEING_TESTED,
+            searchEventAndCase,
+            authorizationHeadersProvider.getLawFirmAAuthorization()
+        );
+
+        result.then().assertThat()
+            .statusCode(HttpStatus.FORBIDDEN.value())
+            .contentType(APPLICATION_JSON_VALUE)
+            .body("timestamp", lessThanOrEqualTo(LocalDateTime.now()
+                                                     .format(DateTimeFormatter.ofPattern(DATE_TIME_FORMAT))))
+            .body("error", equalTo(HttpStatus.FORBIDDEN.getReasonPhrase()))
+            .body("status", equalTo(HttpStatus.FORBIDDEN.value()))
+            .body("message", equalTo("User did not have sufficient permissions to perform this action"));
+    }
+
+    @Test
+    public void should_return_a_200_and_retrieve_a_task_by_event_and_case_location_match() {
+
+        Map<CamundaVariableDefinition, String> variablesOverride = Map.of(
+            CamundaVariableDefinition.JURISDICTION, "IA",
+            CamundaVariableDefinition.LOCATION, "765324",
+            CamundaVariableDefinition.TYPE, "ReviewTheAppeal",
+            CamundaVariableDefinition.TASK_ID, "ReviewTheAppeal",
+            CamundaVariableDefinition.TASK_STATE, "unassigned"
+        );
+
+        Map<String, String> task = common.setupTaskAndRetrieveIdsWithCustomVariablesOverride(variablesOverride);
+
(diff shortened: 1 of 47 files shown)


src/testUtils/java/uk/gov/hmcts/reform/wataskmanagementapi/config/RestApiActions.java
```diff
@@ -27,6 +28,7 @@ public RestApiActions(final String baseUri, final PropertyNamingStrategy propert
     }
 
     public RestApiActions setUp() {
+        port = 8090;
```
iguannago, line 31: can we inject this port rather than hardcode it?

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/TaskController.java
```diff
@@ -289,4 +290,49 @@ public TaskController(CamundaService camundaService,
...
+        @ApiResponse(
+            code = 415,
+            message = "Unsupported Media Type"
+        ),
+        @ApiResponse(
+            code = 500,
+            message = "Internal Server Error"
+        )
+    })
+    @PostMapping(path = "/auto-complete", produces = APPLICATION_JSON_VALUE)
```
raghera, line 319: We should discuss this endpoint.  The naming is quite important as this is what ExUI will use and become part of the specification.

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskForSearchCompletionControllerTest.java
```diff
@@ -0,0 +1,107 @@
...
+            ENDPOINT_BEING_TESTED,
+            searchEventAndCase,
+            authorizationHeadersProvider.getTribunalCaseworkerAAuthorization()
+        );
+
+        result.then().assertThat()
+            .statusCode(HttpStatus.OK.value())
+            .contentType(APPLICATION_JSON_VALUE)
+            .and().contentType(MediaType.APPLICATION_JSON_VALUE);
+
```
iguannago, line 61: Maybe better if you check the response matches the task you are expecting.

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskForSearchCompletionControllerTest.java
```diff
@@ -0,0 +1,107 @@
...
+        SearchEventAndCase searchEventAndCase = new SearchEventAndCase(task.get("caseId"), "requestRespondentEvidence");
+
+        Response result = restApiActions.post(
+            ENDPOINT_BEING_TESTED,
+            searchEventAndCase,
+            authorizationHeadersProvider.getTribunalCaseworkerBAuthorization()
+        );
+
+        result.then().assertThat()
```
iguannago, line 77: Same as above. It would be more robust if you check the expected task is the one return in the response.

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskForSearchCompletionControllerTest.java
```diff
@@ -0,0 +1,107 @@
...
+            .statusCode(HttpStatus.OK.value())
+            .and().contentType(MediaType.APPLICATION_JSON_VALUE);
+    }
+
+    @Test
+    public void should_return_a_500_and_retrieve_a_task_by_event_and_case() {
```
iguannago, line 83: Is the test title correct? If there is a 500 internal error, is the endpoint still returning the task?

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/domain/entities/idam/SearchEventAndCase.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+@ApiModel(
+    value       = "SearchEventAndCase",
+    description = "Search task request with event and case id"
+)
+@SuppressWarnings({"PMD.UnnecessaryAnnotationValueElement"})
```
iguannago, line 11: I believe you can fix this by removing the JsonProperty from either the field or from the getMethod. I don't think you need both.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaService.java
```diff
@@ -175,6 +182,42 @@ public void completeTask(String taskId,
 
     }
 
+    public List<Task> searchForCompletableTasksUsingEventAndCaseId(SearchEventAndCase searchEventAndCase,
+                                                                   List<PermissionTypes> permissionsRequired,
+                                                                   AccessControlResponse accessControlResponse) {
+
+        //1. Perform the DMN evaluation
```
iguannago, line 189: Suggestion: I'd remove these comments. I think the code explains it by itself :) 
Or if you think it's not clear then move the block into a method and choose a convenient name for the method to explain what it does such as: private void evaluateDmn(...)

iguannago (approved):

iguannago (dismissed):

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskForSearchCompletionControllerTest.java
```diff
@@ -0,0 +1,122 @@
...
+        Response result = restApiActions.post(
+            ENDPOINT_BEING_TESTED,
+            searchEventAndCase,
+            authorizationHeadersProvider.getTribunalCaseworkerBAuthorization()
+        );
+
+        result.then().assertThat()
+            .statusCode(HttpStatus.OK.value())
+            .contentType(APPLICATION_JSON_VALUE)
+            .body("tasks[0]", equalTo(null));
```
BenNzewi, line 92: The test name doesn't appear to be in line with the outcome.

If the test is searching/retrieving by retrieve_a_task_by_id_jurisdiction_location_and_region_match, should the task returned be null?

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskForSearchCompletionControllerTest.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
```
BenNzewi, line 94: There are 3 other AC's that are not covered. Can this be added to the functional test;
Given the consuming service is completing an event,
And the consuming service provides the ID of the event that is being completed and the case ID;

AC1
And the event ID matches a row (or rows) in the Task Completion DMN table
And the matching task types are found in the active tasks for that Case
Then return a list of tasks that are completed by that event
And this list should include of the standard attributes as defined above 

AC2
And the event ID does not match any row in the Task Completion DMN table
Then return an empty response, indicating no matches found

AC3
And the event ID matches a row (or rows) in the Task Completion DMN table
And the matching task types are NOT found in the active tasks for that Case
Then return an empty response, indicating no matches found

BenNzewi (approved):

PR #93, opened 2020-12-03
Added DB Tasks cleanup after each test

indvamshi:

https://tools.hmcts.net/jira/browse/RWA-219

### Change description ###

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/GetTaskByIdControllerTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/GetTaskByIdControllerTest.java
@@ -1,7 +1,6 @@
 package uk.gov.hmcts.reform.wataskmanagementapi.controllers;
 
 import io.restassured.response.Response;
-import org.junit.After;
 import org.junit.Test;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
@@ -22,12 +21,10 @@ import static uk.gov.hmcts.reform.wataskmanagementapi.domain.entities.camunda.Ca
 public class GetTaskByIdControllerTest extends SpringBootFunctionalBaseTest {
 
     private static final String ENDPOINT_BEING_TESTED = "task/{task-id}";
-    private String taskId;
 
     @Test
     public void should_return_a_404_if_task_does_not_exist() {
         String nonExistentTaskId = "00000000-0000-0000-0000-000000000000";
-        taskId = nonExistentTaskId;
 
         Response result = restApiActions.get(
             ENDPOINT_BEING_TESTED,
@@ -53,7 +50,7 @@ public class GetTaskByIdControllerTest extends SpringBootFunctionalBaseTest {
     public void should_return_a_401_when_the_user_did_not_have_any_roles() {
 
         Map<String, String> task = common.setupTaskAndRetrieveIds(Common.TRIBUNAL_CASEWORKER_PERMISSIONS);
-        taskId = task.get("taskId");
+        var taskId = task.get("taskId");
 
         Response result = restApiActions.get(
             ENDPOINT_BEING_TESTED,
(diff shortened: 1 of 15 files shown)


src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/GetTaskByIdControllerTest.java
```diff
@@ -21,14 +23,16 @@
 public class GetTaskByIdControllerTest extends SpringBootFunctionalBaseTest {
 
     private static final String ENDPOINT_BEING_TESTED = "task/{task-id}";
+    private static final String ENDPOINT_COMPLETE_TASK = "task/{task-id}/complete";
+    private String taskId;
 
     @Test
     public void should_return_a_404_if_task_does_not_exist() {
-        String nonExistentTaskId = "00000000-0000-0000-0000-000000000000";
+        taskId = "00000000-0000-0000-0000-000000000000";
```
iguannago, line 31: Personal opinion, I'd rather use the "nonExistentTaskId" since it describes more the intention and it's easier to understand what's going on. As I say, this is just a suggestion. I am happy with this change.

iguannago (approved):

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/GetTaskByIdControllerTest.java
indvamshi: I agree. I had to change the label for reuse.

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskAssignByIdControllerTest.java
```diff
@@ -150,5 +156,10 @@ public void should_return_a_403_when_the_assignee_does_not_have_execute_or_own_p
                   equalTo("User did not have sufficient permissions to assign task with id: " + taskId));
     }
 
+    @After
+    public void cleanUp() {
+        camundaApiActions.post(ENDPOINT_COMPLETE_TASK, taskId,
+                               new Headers(authorizationHeadersProvider.getServiceAuthorizationHeader()));
```
alexbaeza, line 162: could we have something like:
instead of wrapping it in new Headers `authorizationHeadersProvider.getServiceAuthorizationHeadersOnly()`

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/GetTaskByIdControllerTest.java
alexbaeza: Not to be fussy, but I can see why you've done it like that in order to clean up the task in camunda however for this one we are not creating a task so I think I also agree with David on having a more descriptive intention :D

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskForSearchCompletionControllerTest.java
```diff
@@ -133,5 +133,11 @@ public void should_return_a_200_and_when_performing_search_when_caseId_correct_e
             .contentType(APPLICATION_JSON_VALUE)
             .body("tasks.size()", equalTo(0));
     }
+
+    @After
+    public void cleanUp() {
+        camundaApiActions.post(ENDPOINT_COMPLETE_TASK, taskId,
+                               new Headers(authorizationHeadersProvider.getServiceAuthorizationHeader()));
+    }
```
alexbaeza, line 141: Again more of personal opinion but would it be better to move this to Common class inside the testUtils. mainly because this forces you to define ENDPOINT_COMPLETE_TASK in this class which is not really relevant to what the class is testing. 

That way we could have something like.

```
common.cleanUpTask(taskId)
```
or
```
common.tearDownTask(taskId)
```

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/GetTaskByIdControllerTest.java
indvamshi: The tearDown test uses taskId, in this case it will be null so completeTask endpoint fails. Alternatively I can do a null check in teardown for taskId but I feel it's unnecessary. IMHO, the test name describes what is tested. There is a tradeoff which way to go.

iguannago (approved):

alexbaeza (approved):

alexbaeza (approved):

src/testUtils/java/uk/gov/hmcts/reform/wataskmanagementapi/utils/Common.java
```diff
@@ -113,5 +120,10 @@ public void updateTaskWithCustomVariablesOverride(Map<String,String> task,
 
     }
 
+    public void cleanUpTask(String taskId) {
+        camundaApiActions.post(ENDPOINT_COMPLETE_TASK, taskId,
+                               authorizationHeadersProvider.getServiceAuthorizationHeadersOnly());
+    }
```
BenNzewi, line 126: Can we check the state of the completed task? I.e. check that the task no longer exists as part of completing and clean up.

iguannago (approved):

alexbaeza (approved):

PR #100, opened 2020-12-14
Rwa 181

indvamshi:

https://tools.hmcts.net/jira/browse/RWA-181
### Change description ###

```
[x] Yes
[ ] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/GetTaskByIdControllerTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/GetTaskByIdControllerTest.java
@@ -17,6 +17,7 @@ import static org.hamcrest.Matchers.lessThanOrEqualTo;
 import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
 import static uk.gov.hmcts.reform.wataskmanagementapi.domain.entities.camunda.CamundaVariableDefinition.JURISDICTION;
 import static uk.gov.hmcts.reform.wataskmanagementapi.domain.entities.camunda.CamundaVariableDefinition.REGION;
+import static uk.gov.hmcts.reform.wataskmanagementapi.utils.Common.REASON_COMPLETED;
 
 public class GetTaskByIdControllerTest extends SpringBootFunctionalBaseTest {
 
@@ -67,7 +68,7 @@ public class GetTaskByIdControllerTest extends SpringBootFunctionalBaseTest {
             .body("status", equalTo(HttpStatus.UNAUTHORIZED.value()))
             .body("message", equalTo("User did not have sufficient permissions to perform this action"));
 
-        common.cleanUpTask(taskId);
+        common.cleanUpTask(taskId, Common.REASON_COMPLETED);
     }
 
     @Test
@@ -87,7 +88,7 @@ public class GetTaskByIdControllerTest extends SpringBootFunctionalBaseTest {
             .and().contentType(MediaType.APPLICATION_JSON_VALUE)
             .and().body("task.id", equalTo(taskId));
 
-        common.cleanUpTask(taskId);
+        common.cleanUpTask(taskId, REASON_COMPLETED);
     }
 
     @Test
@@ -112,7 +113,7 @@ public class GetTaskByIdControllerTest extends SpringBootFunctionalBaseTest {
                 format("User did not have sufficient permissions to access task with id: %s", taskId)
             ));
 
-        common.cleanUpTask(taskId);
+        common.cleanUpTask(taskId, REASON_COMPLETED);
     }
 
     @Test
@@ -132,7 +133,7 @@ public class GetTaskByIdControllerTest extends SpringBootFunctionalBaseTest {
             .and().contentType(MediaType.APPLICATION_JSON_VALUE)
             .and().body("task.id", equalTo(taskId));
 
-        common.cleanUpTask(taskId);
+        common.cleanUpTask(taskId, REASON_COMPLETED);
     }
 
     @Test
@@ -157,7 +158,7 @@ public class GetTaskByIdControllerTest extends SpringBootFunctionalBaseTest {
                 format("User did not have sufficient permissions to access task with id: %s", taskId)
             ));
 
-        common.cleanUpTask(taskId);
+        common.cleanUpTask(taskId, REASON_COMPLETED);
     }
 
 }
--- a/src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskCancelByIdControllerTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskCancelByIdControllerTest.java
@@ -1,7 +1,5 @@
 package uk.gov.hmcts.reform.wataskmanagementapi.controllers;
 
-import io.restassured.http.Header;
-import io.restassured.http.Headers;
 import io.restassured.response.Response;
 import org.junit.Test;
 import org.springframework.http.HttpStatus;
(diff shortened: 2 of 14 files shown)


src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskCancelByIdControllerTest.java
```diff
@@ -0,0 +1,115 @@
...
+    public void should_return_a_204_when_cancelling_a_task_by_id() {
+        Map<String, String> task = common.setupTaskAndRetrieveIds(Common.TRIBUNAL_CASEWORKER_PERMISSIONS);
+        var taskId = task.get("taskId");
+
+        Response result = restApiActions.post(
+            ENDPOINT_BEING_TESTED,
+            taskId,
+            authorizationHeadersProvider.getTribunalCaseworkerAAuthorization()
+        );
+
```
iguannago, line 108: shall we check that the taskState is completed or delete or whatever is expected?

lewis150193 (approved):

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/GetTaskByIdControllerTest.java
```diff
@@ -67,7 +68,7 @@ public void should_return_a_401_when_the_user_did_not_have_any_roles() {
             .body("status", equalTo(HttpStatus.UNAUTHORIZED.value()))
             .body("message", equalTo("User did not have sufficient permissions to perform this action"));
 
-        common.cleanUpTask(taskId);
+        common.cleanUpTask(taskId, Common.REASON_COMPLETED);
```
raghera, line 71: should the Task be Completed in a Cancellation scenario?

iguannago (approved):

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/GetTaskByIdControllerTest.java
indvamshi: Here we are validating against history table which is the Camunda state 'completed'

raghera: I thought it was `Deleted` when you showed me on Termination?

raghera: sorry - ignore that.  This is a 401 scenario

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaService.java
```diff
@@ -439,6 +460,19 @@ private void updateTaskStateTo(String taskId, TaskState newState) {
         camundaServiceApi.addLocalVariablesToTask(authTokenGenerator.generate(), taskId, camundaLocalVariables);
     }
 
+    private void performCancelTaskAction(String taskId) {
+        Map<String, String> body = new ConcurrentHashMap<>();
+        body.put("escalationCode", ESCALATION_CODE);
+        try {
+            camundaServiceApi.bpmnEscalation(authTokenGenerator.generate(), taskId, body);
+        } catch (FeignException ex) {
```
raghera, line 468: Could we use something like Feign ErrorDecoder or ControllerAdvice to help keep the code cleaner?  Maybe something to think about as a future improvement across all apis?

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/TaskControllerTest.java
```diff
@@ -160,4 +160,12 @@ void should_auto_complete_a_task() {
         assertNotNull(response);
         assertEquals(HttpStatus.OK, response.getStatusCode());
     }
+
+    @Test
+    void should_cancel_a_task() {
```
raghera, line 165: Can we add some exception scenarios?

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaServiceTest.java
```diff
@@ -1085,5 +1086,137 @@ void should_auto_complete_and_return_empty_array() {
...
+                .isInstanceOf(ServerErrorException.class)
+                .hasCauseInstanceOf(FeignException.class)
+                .hasMessage(String.format(
+                    "There was a problem cancelling the task with id: %s",
+                    taskId));
+
+        }
+
+    }
+
```
BenNzewi, line 1220: do we need a test scenario for what happens when a task we are trying to cancel does not exist?

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskCancelByIdControllerTest.java
BenNzewi, line 108: AC2 will also need to be covered. It's similar to other endpoints. I.e. testing CASE ROLE/CASE ID after we set up/mock full ccd integration. (We can discuss- it will be a separate ticket).

AC5 -> Then the endpoint will return a 5XX range HTTP error (role assignment service error); This can be in an integration test. We can also discuss.

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaServiceTest.java
indvamshi: Its covered in FT, have a look at 404 test.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaService.java
indvamshi: yes we have to use Rest exception handler or ErrorDecoder to handle FeignExceptions. 
We can create a tech debt story for this.

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/TaskControllerTest.java
indvamshi: This method does not throw any exceptions, the validations are done in service layer and the tests are written to handle all exceptions.

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskCancelByIdControllerTest.java
indvamshi: AC2 - yes it can be part of the ticket 225.
AC5 - I think we got a story for writing integration tests to handle 500, so this can be part of that ticket.

indvamshi: AC5- new ticket created. 
https://tools.hmcts.net/jira/browse/RWA-281

raghera: As a general principle, we should cover all the ACs as part of the feature ticket as this should be incorporated into the points. Otherwise, if tests are written after the feature is merged, there is an opportunity for a bug to be merged.  

In this case we can do the work for 225 separately since the Jira is already there.  281 should have been done here so maybe merge this and work on that next.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaService.java
raghera: Yes, create a tech debt ticket for that please.

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskCancelByIdControllerTest.java
indvamshi: sure.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaService.java
indvamshi: https://tools.hmcts.net/jira/browse/RWA-282

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/TaskControllerTest.java
raghera: Just trying to point out that we should think about the failure scenarios in general.  Could verify for instance that the taskId you pass in is the same one used in cancelling the Task.

Also, If the role assignment service is down and throws an unchecked exception there could be a test around that.  Even if the test shows that no handling is required, if in future someone adds code to handle it differently the existing test would break.

PR #109, opened 2021-01-04
Rwa 230

indvamshi:

https://tools.hmcts.net/jira/browse/RWA-230

### Change description ###

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskForSearchCompletionControllerTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskForSearchCompletionControllerTest.java
@@ -19,15 +19,15 @@ import static uk.gov.hmcts.reform.wataskmanagementapi.utils.Common.REASON_COMPLE
 
 public class PostTaskForSearchCompletionControllerTest extends SpringBootFunctionalBaseTest {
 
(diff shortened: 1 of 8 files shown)


lewis150193 (approved):

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskForSearchCompletionControllerTest.java
```diff
@@ -142,5 +150,45 @@ public void should_return_a_200_and_when_performing_search_when_caseId_correct_e
         common.cleanUpTask(taskId, REASON_COMPLETED);
     }
 
+    @Test
```
BenNzewi, line 153: Excellent scenario coverage.

One small suggestion; So we adhere to the test pyramid and lean coverage in FT(One positive test case + 1 other edge case), could you move "should_return_a_500_and_when_performing_search_when_caseType_is_incorrect()" under integration test?

indvamshi: The reason for adding jurisdiction/casetype in FT is it actually calls camunda endpoint and checks for both the keys. But in IT, we are mocking to respond to 500 errorCode so the key is irrelevant.

PR #130, opened 2021-02-09
Added waits

alexbaeza:
### Change description ###

Fix pipeline
Fixes tests
Improve Logging
Added polling mechanism

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/auth/permission/PermissionEvaluatorService.java
```diff
@@ -65,43 +67,52 @@ private boolean evaluateAccess(Map<String, CamundaVariable> variables,
                                    List<PermissionTypes> permissionsRequired) {
         boolean hasAccess;
 
+        log.debug("Evaluating access for {}", roleAssignment);
```
iguannago, line 70: I think it would be good to turn Debug mode ON in AAT, so that we can see these logs.

indvamshi (approved):

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/auth/permission/PermissionEvaluatorService.java
alexbaeza: yes this would be amazing :)

indvamshi: If it is not on, we should then to trouble shoot.

PR #152, opened 2021-02-22
RWA-374: Search optimization

alexbaeza:

https://tools.hmcts.net/jira/browse/RWA-374
### Change description ###

Optimize search queries by using a different implementation when retrieving variables.

Notes: Instead of retrieving the task variables in n+1 calls we optimized this call by using the variable-instance endpoint in camunda and passing in a list of process variables to retrieve all variables in one shot. 
This however means that we need to do some post-processing to associate the variables to the task, but optimizes our search by almost 300%.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaServiceTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaServiceTest.java
@@ -384,6 +384,57 @@ class CamundaServiceTest extends CamundaServiceBaseTest {
             verifyNoMoreInteractions(camundaServiceApi);
         }
 
+
+        @Test
+        void searchWithCriteria_should_succeed_and_return_empty_list_if_task_had_no_variables() {
+            List<Assignment> roleAssignment = singletonList(mock(Assignment.class));
+            List<PermissionTypes> permissionsRequired = singletonList(READ);
+            SearchTaskRequest searchTaskRequest = mock(SearchTaskRequest.class);
+            CamundaSearchQuery camundaSearchQueryMock = mock(CamundaSearchQuery.class);
+            ZonedDateTime dueDate = ZonedDateTime.now().plusDays(1);
+            CamundaTask camundaTask = new CamundaTask(
+                "someId",
+                "someTaskName",
+                "someAssignee",
+                ZonedDateTime.now(),
+                dueDate,
+                null,
+                null,
+                "someFormKey",
+                "someProcessInstanceId"
+            );
+
+            when(camundaQueryBuilder.createQuery(searchTaskRequest))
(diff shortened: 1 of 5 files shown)


iguannago (dismissed):

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaServiceTest.java
```diff
@@ -251,19 +269,19 @@ void searchWithCriteria_should_succeed() {
...
             when(camundaQueryBuilder.createQuery(searchTaskRequest))
                 .thenReturn(camundaSearchQueryMock);
             when(camundaServiceApi.searchWithCriteria(BEARER_SERVICE_TOKEN, camundaSearchQueryMock.getQueries()))
                 .thenReturn(singletonList(camundaTask));
-            when(camundaServiceApi.getVariables(BEARER_SERVICE_TOKEN, camundaTask.getId()))
-                .thenReturn(variables);
+            when(camundaServiceApi.getAllVariables(
+                BEARER_SERVICE_TOKEN,
+                Map.of("variableScopeIdIn", singletonList("someProcessInstanceId"))
+            )).thenReturn(mockedVariablesResponse());
```
raghera, line 279: Not sure if I've missed them elsewhere but probably want some unhappy cases - E.g. if empty response returned?  Also if we have different datasets coming back like when searching for multiple Tasks belonging to different processes maybe?

alexbaeza: You're right, good spot we need to cover those scenarios I will add more unit tests thanks :)

indvamshi (approved):

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaServiceTest.java
```diff
@@ -313,19 +346,227 @@ void searchWithCriteria_should_succeed_and_return_empty_list_if_user_did_not_hav
...
+                camundaSearchQueryMock.getQueries()
+            );
+            verify(camundaServiceApi, times(1))
+                .getAllVariables(
+                    BEARER_SERVICE_TOKEN,
+                    Map.of("variableScopeIdIn", singletonList("someProcessInstanceId"))
+                );
+            verifyNoMoreInteractions(camundaServiceApi);
+        }
+
```
BenNzewi, line 537: Do we need a scenario for should_return_empty_list_when_multiple_tasks_but_no_permissions?

alexbaeza: Added more scenarios as discussed

BenNzewi (approved):

PR #160, opened 2021-03-05
fixed 500 and refactoring of endpoint name and snake-casing

lewis150193:
### Change description ###

```
[ ] Yes
[ x ] No
```


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaService.java
```diff
@@ -239,20 +239,30 @@ public void completeTask(String taskId,
...
+               searchEventAndCase.getCaseType().equals("asylum")) {
+                // A List (Array) with a map (One object) with objects inside the object (String and CamundaVariable).
+               evaluateDmnResult =
+                    camundaServiceApi.evaluateDMN(
+                        authTokenGenerator.generate(),
+                        getTableKey(searchEventAndCase.getCaseJurisdiction(), searchEventAndCase.getCaseType()),
+                        dmnRequest
+                    );
+            } else {
+                return Collections.emptyList();
```
raghera, line 253: I’m not sure returning an empty list is the right response if the request is sending effectively invalid data.  I think for MVP we can say that we only accept the jurisdiction of ia and caseType or asylum .  So I think the response should be an http 400 Bad Request response with text explaining what is wrong.

The text can be: ```Please check your request.  This endpoint currently only supports the Immigration & Asylum service.```

raghera: Should add a test for this scenario too.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaService.java
```diff
@@ -239,20 +239,30 @@ public void completeTask(String taskId,
...
+            }
 
             List<String> taskTypes = evaluateDmnResult.stream()
                 .map(result -> getVariableValue(result.get("task_type"), String.class))
                 .collect(Collectors.toList());
 
 
-            if (taskTypes.isEmpty()) {
+            if (taskTypes.isEmpty() || evaluateDmnResult == null) {
```
raghera, line 265: Not sure if the `evaluateDmnResult` should be null here?  You are already calling `evaluateDmnResult.isEmpty()` in line 256 which would be an NPE if it's null?

Ideally we don't want any nulls in the code.  This is a List so if there are no results it should an empty list.

src/main/resources/application.yaml
```diff
@@ -56,7 +56,7 @@ idam:
   s2s-authorised:
     services: ${WA_S2S_AUTHORIZED_SERVICES:ccd,ccd_data,ccd_gw,ccd_ps,iac,wa_task_management_api,xui_webapp}
   api:
-    baseUrl: ${IDAM_API_URL:http://sidam-api}
+    baseUrl: ${IDAM_API_URL:http://sidam-simulator}
```
raghera, line 59: Is this deliberate?  This seems like a change for development only?

raghera (changes requested):

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskForSearchCompletionControllerTest.java
```diff
@@ -121,6 +121,29 @@ public void should_return_a_200_and_return_and_empty_list_when_event_id_does_not
...
+
+        Response result = restApiActions.post(
+            ENDPOINT_BEING_TESTED,
+            searchEventAndCase,
+            authenticationHeaders
+        );
+
+        result.then().assertThat()
+            .statusCode(HttpStatus.BAD_REQUEST.value())
+            .contentType(APPLICATION_JSON_VALUE);
```
raghera, line 142: could check the error string coming back to the client is correct too.

raghera (dismissed):

raghera (approved):

PR #170, opened 2021-03-17
Bugfix/rwa 415 base location

iguannago:
https://tools.hmcts.net/jira/browse/RWA-415

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/auth/role/entities/RoleAttributeDefinition.java
+++ b/src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/auth/role/entities/RoleAttributeDefinition.java
@@ -5,6 +5,7 @@ public enum RoleAttributeDefinition {
     CASE_ID("caseId"),
     JURISDICTION("jurisdiction"),
     PRIMARY_LOCATION("primaryLocation"),
+    BASE_LOCATION("baseLocation"),
     CASE_TYPE("caseType"),
     REGION("region");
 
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/auth/role/entities/RoleAttributeDefinition.java
```diff
@@ -5,6 +5,7 @@
     CASE_ID("caseId"),
     JURISDICTION("jurisdiction"),
     PRIMARY_LOCATION("primaryLocation"),
```
raghera, line 7: Just wondered if this enum a representation of any Role Assignment Attributes or just the ones we care about for Access Control?  Just considering if we should remove primaryLocation.

raghera (approved):

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/auth/role/entities/RoleAttributeDefinition.java
iguannago: yeah, based on the roleAssignment service: https://github.com/hmcts/am-role-assignment-service/blob/c838cea241e94d452d528a0f2fe653305939c02b/src/main/java/uk/gov/hmcts/reform/roleassignment/domain/model/Assignment.java#L25 

The attributes field is a free json object and so anyone can put in there what they want, it looks like. 

I think to be safe we could put a default value: @JsonEnumDefaultValue UNKNOWN

PR #182, opened 2021-03-31
Added intetgration test for restricted role assignment

indvamshi:

https://tools.hmcts.net/jira/browse/RWA-411

### Change description ###

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskSearchControllerTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskSearchControllerTest.java
@@ -3,7 +3,6 @@ package uk.gov.hmcts.reform.wataskmanagementapi.controllers;
 import io.restassured.http.Headers;
 import io.restassured.response.Response;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.springframework.http.HttpStatus;
 import uk.gov.hmcts.reform.wataskmanagementapi.SpringBootFunctionalBaseTest;
@@ -26,6 +25,7 @@ import static java.util.Collections.singletonList;
 import static org.hamcrest.CoreMatchers.hasItems;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.Matchers.either;
+import static org.hamcrest.Matchers.empty;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.everyItem;
 import static org.hamcrest.Matchers.hasItem;
@@ -167,13 +167,12 @@ public class PostTaskSearchControllerTest extends SpringBootFunctionalBaseTest {
         common.cleanUpTask(taskId, REASON_COMPLETED);
     }
 
-    @Ignore("https://tools.hmcts.net/jira/browse/RWA-411")
     @Test
     public void should_return_a_200_with_search_results_with_restricted_role_assignment() {
         TestVariables taskVariablesSscs =
             common.setupTaskAndRetrieveIdsWithCustomVariable(CamundaVariableDefinition.JURISDICTION, "SSCS");
-        String taskId = taskVariablesSscs.getTaskId();
-        String caseId = taskVariablesSscs.getCaseId();
+        final String taskId = taskVariablesSscs.getTaskId();
+        final String caseId = taskVariablesSscs.getCaseId();
 
         common.setupTaskAndRetrieveIdsWithCustomVariable(CamundaVariableDefinition.JURISDICTION, "IA");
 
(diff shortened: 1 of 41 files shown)


lewis150193 (approved):

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskSearchControllerTest.java
```diff
@@ -192,9 +191,7 @@ public void should_return_a_200_with_search_results_with_restricted_role_assignm
 
         result.then().assertThat()
             .statusCode(HttpStatus.OK.value())
-            .body("tasks.jurisdiction", everyItem(is("SSCS")))
-            .body("tasks.case_id", hasItem(caseId))
-            .body("tasks.id", hasItem(taskId));
+            .body("tasks", empty());
```
alexbaeza, line 194: I don't think this is correct. 

This test was disabled because it is just not possible to do through functional tests unless we create a real SSCS case, due to the task configuration configuring the task and adding the correct jurisdiction and case type, Hence it must be tested as part of an integration test.

So the assertion is wrong here the one that was disabled previously is the correct one however if this particular scenario has been covered by an integration tests then we can just remove this functional test.

src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/TaskControllerTest.java
```diff
@@ -128,6 +142,68 @@ void should_return_a_500_when_id_invalid() throws Exception {
...
+            mockCamundaVariables();
+
+            mockMvc.perform(
+                get("/task/" + taskId)
+                    .header(
+                        "Authorization",
+                        authorizationHeadersProvider.getTribunalCaseworkerAAuthorization()
+                    )
+                    .contentType(MediaType.APPLICATION_JSON_VALUE)
+            ).andExpect(status().is4xxClientError());
```
alexbaeza, line 192: we should probably change this to be more specific I believe in this particular scenario we want this to fail as the user didn't have enough permissions to the task as he was in a different jurisdiction so in this case we expect a 403.

src/testUtils/java/uk/gov/hmcts/reform/wataskmanagementapi/config/GivensBuilder.java
```diff
@@ -114,7 +114,7 @@ public GivensBuilder iCreateATaskWithCaseId(String caseId) {
         AtomicReference<List<CamundaTask>> response = new AtomicReference<>();
         await().ignoreException(AssertionError.class)
             .pollInterval(500, MILLISECONDS)
-            .atMost(30, SECONDS)
+            .atMost(60, SECONDS)
```
alexbaeza, line 117: 30 should be enough

alexbaeza (changes requested):
We also need a test that will return a 200 when a user had a role assignment jurisdiction of IA. and a restricted role-assignment for a particular case with jurisdiction SSCS. and that should return a 200 with the task info.

src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/TaskControllerTest.java
```diff
@@ -128,6 +142,68 @@ void should_return_a_500_when_id_invalid() throws Exception {
...
         }
+
+
+        @Test
+        void should_return_a_400_when_restricted_role_is_given() throws Exception {
+            final var taskId = UUID.randomUUID().toString();
+
+            final var userToken = "user_token";
+
+            mockServices.mockUserInfo(idamWebApi);
```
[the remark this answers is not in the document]
indvamshi: This method call is required to mock IdamWebAPI userInfo call

src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/TaskControllerTest.java
indvamshi: Agreed.

src/testUtils/java/uk/gov/hmcts/reform/wataskmanagementapi/config/GivensBuilder.java
indvamshi: will revert.

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskSearchControllerTest.java
indvamshi: The TaskSearch does not throw any 403 if there is any mismatch in permissions instead it throws empty list. 
I agree on creating a case with SCSS if its not supported then it should not be part of any FT.

alexbaeza (approved):

PR #191, opened 2021-04-13
get task pact provider test

lewis150193:
https://tools.hmcts.net/jira/browse/RWA-444

### Change description ###
Provider test task/{taskId}

```
[ ] Yes
[ x ] No
```


iguannago (changes requested):

Jenkinsfile_CNP
```diff
@@ -45,6 +45,9 @@ withPipeline(type, product, component) {
     env.CCD_URL = "http://ccd-data-store-api-aat.service.core-compute-aat.internal"
     env.ROLE_ASSIGNMENT_URL = "http://am-role-assignment-service-aat.service.core-compute-aat.internal"
     env.PACT_BROKER_FULL_URL = "https://pact-broker.platform.hmcts.net"
+    env.PACT_BROKER_URL = "pact-broker.platform.hmcts.net"
+    env.PACT_BROKER_PORT = "443"
+    env.PACT_BROKER_SCHEME = "https"
```
alexbaeza, line 50: do you need these or can we use the PACT_BROKER_FULL_URL instead ?

Jenkinsfile_CNP
```diff
@@ -63,9 +66,11 @@ withPipeline(type, product, component) {
 
     onMaster() {
         enablePactAs([AppPipelineDsl.PactRoles.CONSUMER])
+        enablePactAs([AppPipelineDsl.PactRoles.PROVIDER])
     }
 
     onPR() {
         enablePactAs([AppPipelineDsl.PactRoles.CONSUMER])
+        enablePactAs([AppPipelineDsl.PactRoles.PROVIDER])
```
alexbaeza, line 74: we probably don't want to enable provider tests on or level as we only want to be verifying with what's on master not experimental branches :)

charts/wa-task-management-api/values.yaml
```diff
@@ -18,4 +18,6 @@ java:
     IDAM_API_URL: "https://idam-api.{{ .Values.global.environment }}.platform.hmcts.net"
     ROLE_ASSIGNMENT_URL: "http://am-role-assignment-service-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal"
     CCD_URL: "http://ccd-data-store-api-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal"
-
+    PACT_BROKER_URL: pact-broker.platform.hmcts.net
+    PACT_BROKER_PORT: 443
+    PACT_BROKER_SCHEME: https
```
alexbaeza, line 23: don't think we need these here since the contract tests happen on the pipeline

src/contractTest/java/uk/gov/hmcts/reform/wataskmanagementapi/provider/service/TaskManagementGetTaskByIdPactTest.java
```diff
@@ -0,0 +1,98 @@
...
+                         ZonedDateTime.now(),
+                         ZonedDateTime.now(),
+                         "TCW",
+                         true,
+                         "Time extension",
+                         "ia",
+                         "region",
+                         "12345",
+                         "Newcastle",
+                         "asylum",
```
alexbaeza, line 90: we should double check these values to make sure they are close to similar values e.g jurisdiction and caseType are keys so they should use the correct casing

indvamshi (dismissed):

src/contractTest/java/uk/gov/hmcts/reform/wataskmanagementapi/provider/service/TaskManagementGetTaskByIdPactTest.java
```diff
@@ -0,0 +1,98 @@
...
+        if (context != null) {
+            context.setTarget(testTarget);
+        }
+
+    }
+
+    @State({"appropriate task is returned"})
+    public void getTaskById() {
+        setInitiMock();
+    }
```
alexbaeza, line 67: I'm not enterily sure how yet but we should also make sure they are sending in valid authentication headers e.g
Given a caseworker user exists 
And I call from a whitelisted service 
Then the targetted task is returned. 

This will help when writing some contract tests for 403 and 401 errors if we need to cover those

alexbaeza (changes requested):

build.gradle
```diff
@@ -334,6 +327,14 @@ dependencies {
   contractTestCompile group: 'au.com.dius.pact.consumer', name: 'junit5', version: versions.pactVersion
   contractTestCompile group: 'au.com.dius.pact.consumer', name: 'java8', version: versions.pactVersion
 
+
+  contractTestCompile group: 'au.com.dius.pact.provider', name: 'junit5', version: '4.1.14'
+  contractTestCompile group: 'au.com.dius.pact.provider', name: 'spring', version: '4.1.14'
+  contractTestCompile group: 'au.com.dius.pact.provider', name: 'junit5spring', version: '4.1.14'
+
+  testImplementation "au.com.dius.pact.provider:junit5:4.1.7"
+  testImplementation "au.com.dius.pact.provider:junit5spring:4.1.7"
```
alexbaeza, line 336: we seem to be using different versions here in any case:
we should use the variable defined as versions.pactVersion which points to 4.1.7

build.gradle
```diff
@@ -20,7 +20,7 @@ plugins {
   id 'info.solidsoft.pitest' version '1.5.2'
   id 'uk.gov.hmcts.java' version '0.12.5'
   id 'io.freefair.lombok' version '5.3.0'
-  id 'au.com.dius.pact' version '4.1.0'
+  id 'au.com.dius.pact' version '4.1.14'
```
alexbaeza, line 23: Could you check which is the latest version the plugin might be different but the libraries are on 4.1.7

charts/wa-task-management-api/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: "1.0"
 description: A Helm chart for wa-task-management-api App
 name: wa-task-management-api
 home: https://github.com/hmcts/wa-task-management-api
-version: 0.0.19
+version: 0.0.20
```
alexbaeza, line 6: No need to bump the version anymore since we are not changing the chart

Jenkinsfile_CNP
```diff
@@ -63,6 +66,7 @@ withPipeline(type, product, component) {
 
     onMaster() {
         enablePactAs([AppPipelineDsl.PactRoles.CONSUMER])
+        enablePactAs([AppPipelineDsl.PactRoles.PROVIDER])
     }
 
     onPR() {
```
BenNzewi, line 72: this is required at the PR level as well

src/contractTest/java/uk/gov/hmcts/reform/wataskmanagementapi/provider/service/TaskManagementGetTaskByIdPactTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+    @State({"appropriate task is returned"})
```
BenNzewi, line 71: do these states represent the actual behaviour?

iguannago (dismissed):

PR #210, opened 2021-04-25
RWA-514: Migrate to taskVariables

alexbaeza:

https://tools.hmcts.net/jira/browse/RWA-514

### Change description ###

Correct variables naming strategy to camel case for task completion dmn values
Remove Variable scope Id when retrieving variables for task ids using taskIdIn instead
Migrate from processVariables to taskVariables when searching
Update Test
Update Domain objects
Updated Camunda Query builder to use taskType task variable instead of taskId when searching for completable tasks
Added extra functional tests to assert the correct format of the task object is being returned and values are not null
Minor clean up

```
[ ] Yes
[ ] No
```


src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaServiceTest.java
```diff
@@ -144,7 +143,7 @@ private void verifyTaskStateUpdateWasCalled(String taskId, TaskState newTaskStat
         );
     }
 
-    private List<CamundaVariableInstance> mockedVariablesResponse(String processInstanceId) {
+    private List<CamundaVariableInstance> mockedVariablesResponse(String processInstanceId, String taskId) {
```
raghera, line 146: Maybe this comment belongs elsewhere and you have done it already.  Could be good idea to maintain a test where you search using processVariables and check what we get back.  So in theory these should stay the same and never be updated during a process.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaService.java
```diff
@@ -408,12 +410,12 @@ private void performClaimTaskAction(String taskId, Map<String, String> body) {
...
-                .map(CamundaTask::getProcessInstanceId)
+            List<String> searchResultsTaskIds = searchResults.stream()
+                .map(CamundaTask::getId)
                 .collect(Collectors.toList());
 
-            //Retrieve all variables for processIds
-            Map<String, List<String>> body = Map.of("variableScopeIdIn", searchResultsProcessIds);
+            //Retrieve all variables for taskIds
+            Map<String, List<String>> body = Map.of("taskIdIn", searchResultsTaskIds);
             List<CamundaVariableInstance> allVariables =
```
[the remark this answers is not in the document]
alexbaeza: Only task variables will get returned from this call and no processVariables as these will search for all variables with a task assigned to it a.k.a task local variables.
Process variables and execution local variables do not have a taskId on them.

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaServiceTest.java
alexbaeza: good idea I will raise another PR and use the /assign logic to update the task id and assert that the taskState processVariable should remain unchanged.

lewis150193 (dismissed):

lewis150193 (approved):

PR #227, opened 2021-05-04
RWA-530: Add support for case id in search

alexbaeza:

https://tools.hmcts.net/jira/browse/RWA-530

### Change description ###

Adds support for caseId in searches

```
[ ] Yes
[ x] No
```


lewis150193 (approved):

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CamundaQueryBuilder.java
```diff
@@ -71,12 +71,18 @@ public CamundaSearchQuery createQuery(SearchTaskRequest searchTaskRequest) {
             searchParametersMap.get(STATE)
         );
 
+        CamundaOrQuery.CamundaOrQueryBuilder caseIdQueries = createTaskVariableQueriesFor(
```
raghera, line 74: There's a question over we should be allowing this search immediately since it is technically a strategic requirement rather than MVP.  I don't see any issues as long as it works tbh.  Might be worth asking in standup.

alexbaeza: Sure let's discuss tomorrow as this will also be needed for the post-deployment, I think as you said technically there's no issue with allowing this but it's there if people want to start using it

PR #229, opened 2021-05-06
configure tasks script

lewis150193:
https://tools.hmcts.net/jira/browse/RWA-520

### Change description ###
Configure task script

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/configure_task.sh
+++ b/configure_task.sh
@@ -13,32 +13,31 @@ serviceToken=$(curl --silent --show-error -X POST \
   ${S2S_URL}/testing-support/lease
 )
 
-echo $serviceToken
-
 # GET TASKS
 response=$(curl --request POST \
   --url http://camunda-api-aat.service.core-compute-aat.internal/engine-rest/task \
   --header 'Content-Type: application/json' \
-  --cookie dtCookie=v_4_srv_2_sn_708A526309E3A599F638913244C50A40_perc_100000_ol_0_mul_1_app-3Aea7c4b59f27d43eb_1_rcs-3Acss_0 \
-  --data '{
-    "orQueries": [
+  --data "{
+    'orQueries': [
         {
-            "processVariables": [
+            'processVariables': [
                 {
-                    "name": "taskState",
-                    "operator": "eq",
-                    "value": "unconfigured"
+                    'name': 'taskState',
+                    'operator': 'eq',
+                    'value': 'unconfigured'
                 }
             ]
         }
     ],
-    "taskDefinitionKey": "processTask",
-    "processDefinitionKey": "wa-task-initiation-ia-asylum"
-}')
+    'createdAfter': ${date +'%Y-%m-%dT%H:%M:%S.000%z'},
+    'taskDefinitionKey': 'processTask',
+    'processDefinitionKey': 'wa-task-initiation-ia-asylum'
+}")
 
 # CREATE FILE WITH TASKS WITH NUMBER OF TASKS
 tmpfile=$(mktemp /tmp/tasks.json)
 echo ${response} > /tmp/tasks.json
+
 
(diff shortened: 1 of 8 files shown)


configure_task.sh
```diff
@@ -0,0 +1,60 @@
...
+
+# CREATE FILE WITH TASKS WITH NUMBER OF TASKS
+tmpfile=$(mktemp /tmp/tasks.json)
+echo ${response} > /tmp/tasks.json
+
+# LOOP OVER ALL IDS
+for ID in $(seq 0 $(jq length /tmp/tasks.json)); do
+ID_RESPONSE=$(cat /tmp/tasks.json | jq -r ".[${ID}].id")
+response=$(curl --request POST \
+  --url http://wa-task-configuration-api-aat.service.core-compute-aat.internal/task/${ID_RESPONSE} \
```
alexbaeza, line 47: I believe the urls should be parametized so we can point to different envs

configure_task.sh
```diff
@@ -0,0 +1,60 @@
...
+  ${S2S_URL}/testing-support/lease
+)
+
+echo $serviceToken
+
+# GET TASKS
+response=$(curl --request POST \
+  --url http://camunda-api-aat.service.core-compute-aat.internal/engine-rest/task \
+  --header 'Content-Type: application/json' \
+  --cookie dtCookie=v_4_srv_2_sn_708A526309E3A599F638913244C50A40_perc_100000_ol_0_mul_1_app-3Aea7c4b59f27d43eb_1_rcs-3Acss_0 \
```
alexbaeza, line 22: 🍪  can remove

configure_task.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+## Usage: ./camunda-deployment [SERVICE_TOKEN]
+##
+## Options:
+##    - SERVICE_TOKEN: a service token for a whitelisted service in camunda which is generated with the idam-service-token.
+##
```
alexbaeza, line 6: This should be updated with the correct info :)

configure_task.sh
```diff
@@ -0,0 +1,60 @@
...
+        {
+            "processVariables": [
+                {
+                    "name": "taskState",
+                    "operator": "eq",
+                    "value": "unconfigured"
+                }
+            ]
+        }
+    ],
```
raghera, line 34: I think we should also filter on the Task's age.  Otherwise you could start interfering with Tasks that are _currently_ being configured and try to configure them again.  So maybe the query can look for Tasks that are > 1 min old?  That should give enough time.

configure_task.sh
```diff
@@ -0,0 +1,60 @@
...
+response=$(curl --request POST \
+  --url http://camunda-api-aat.service.core-compute-aat.internal/engine-rest/task \
+  --header 'Content-Type: application/json' \
+  --cookie dtCookie=v_4_srv_2_sn_708A526309E3A599F638913244C50A40_perc_100000_ol_0_mul_1_app-3Aea7c4b59f27d43eb_1_rcs-3Acss_0 \
+  --data '{
+    "orQueries": [
+        {
+            "processVariables": [
+                {
+                    "name": "taskState",
```
raghera, line 28: I think we may have moved this to be a taskVariable not a process one?

raghera: Also worth noting that there will be 2 variables - one with the process scope and one with the Task scope.  Most of the process ones will be unconfigured but at the Task scope they are indeed configured.  Worth understanding the difference.

raghera (changes requested):

configure_task.sh
lewis150193: Would we be able to catch up about this?

PR #266, opened 2021-06-21
RWA-587: Added Task db, schema and migrations

alexbaeza:
### JIRA link (if applicable) ###

https://tools.hmcts.net/jira/browse/RWA-587

### Change description ###

Add new postgres 11 database through terraform configuration
Added secrets to env vars in values.yaml, jenkins and bootstrap.yaml
Added dependencies and preview chart
Added initial schema 

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x ] No
```

PR #267, opened 2021-06-22
RWA-587: Add new postgres 11 database through terraform configuration

alexbaeza:
### Change description ###

Add new postgres 11 database through terraform configuration

```
[ ] Yes
[x ] No
```


infrastructure/variables.tf
```diff
@@ -19,3 +19,13 @@ variable "common_tags" {
 variable "appinsights_instrumentation_key" {
   default = ""
 }
+
+
+variable "postgresql_database_name" {
+  default = "wa_task_management_api"
```
raghera, line 25: I know this is just the default but wondered if we want to call it something else?  `cft-task-db` ?  To match with the domain language?

alexbaeza: sure will apply the change

infrastructure/variables.tf
```diff
@@ -19,3 +19,15 @@ variable "common_tags" {
 variable "appinsights_instrumentation_key" {
   default = ""
 }
+
+variable "postgres_db_component_name" {
```
iguannago, line 23: do we need this? the wa-workflow-api does not have it, though

iguannago:
Shall we also add the DB env vars to both the bootstrap and Charts?

infrastructure/variables.tf
alexbaeza: as per suggestion from @raghera we want to name this database host and db differently so I saved it as a variable so it's easier to control.

iguannago (approved):

alexbaeza:
> Shall we also add the DB env vars to both the bootstrap and Charts?

Hey @iguannago This is done as part of the second PR:
https://github.com/hmcts/wa-task-management-api/pull/266

As you can see the second PR fails as I need to create the infrastructure and let terraform add those secrets into the vault however terraform only runs on master, so I have split the PRs into two.

PR #286, opened 2021-07-13
RWA-585 Update task configuration to include case management category

mkayad:

[RWA-585](https://tools.hmcts.net/jira/browse/RWA-585)

### Change description ###

See ticket

```
[ ] Yes
[ X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/domain/entities/camunda/CamundaVariableDefinition.java
+++ b/src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/domain/entities/camunda/CamundaVariableDefinition.java
@@ -22,7 +22,8 @@ public enum CamundaVariableDefinition {
     TITLE("title"),
     HAS_WARNINGS("hasWarnings"),
     WARNING_LIST("warningList"),
-    CFT_TASK_STATE("cftTaskState");
+    CFT_TASK_STATE("cftTaskState"),
+    CASE_MANAGEMENT_CATEGORY("caseManagementCategory");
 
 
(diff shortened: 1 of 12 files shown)


iguannago (dismissed):

iguannago:
there is no test for that change? Do we need to add any?

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/domain/entities/camunda/CamundaVariableDefinition.java
```diff
@@ -21,7 +21,8 @@
     TASK_TYPE("taskType"),
     TITLE("title"),
     HAS_WARNINGS("hasWarnings"),
-    WARNING_LIST("warningList");
+    WARNING_LIST("warningList"),
+    CASE_MANAGEMENT_CATEGORY("caseManagementCategory");
```
alexbaeza, line 26: If the new variable is one of the required variables then we also need to update the tests and the configuration api logic to add it as a local task variable

alexbaeza (changes requested):

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/domain/entities/camunda/CamundaVariableDefinition.java
mkayad: I had more code in the new commit

alexbaeza (approved):

indvamshi (approved):

PR #291, opened 2021-07-14
RWA-589: Completion / Cancellation Lock transaction and rollback

alexbaeza:

https://tools.hmcts.net/jira/browse/RWA-589

### Change description ###

Lock transaction and rollback if something goes wrong in camunda

```
[ ] Yes
[ x] No
```


src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskManagementServiceTest.java
```diff
@@ -0,0 +1,383 @@
...
+        assertEquals(null, savedTaskResource.get().getState());
+
+    }
+
+    private void createAndSaveTestTask(String taskId) {
+        TaskResource taskResource = new TaskResource(taskId, "taskName", "taskType");
+        taskResourceRepository.save(taskResource);
+    }
+
+    @Nested
```
iguannago, line 161: I like the nested thing to organise it better and reuse stuff from the base class. We could also move these nested classes to their own class and extend from this base class if we see the file gets too big over time.

iguannago (approved):

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskManagementService.java
```diff
@@ -170,7 +176,22 @@ public void cancelTask(String taskId,
...
         roleAssignmentVerification(variables, accessControlResponse.getRoleAssignments(), permissionsRequired);
 
-        camundaService.cancelTask(taskId);
+        boolean isFeatureEnabled = launchDarklyFeatureFlagProvider.getBooleanValue(
+            FeatureFlag.RELEASE_2_CANCELLATION_COMPLETION_FEATURE,
+            accessControlResponse.getUserInfo().getUid()
+        );
+
+        if (isFeatureEnabled) {
+            //Lock & update Task
```
iguannago, line 185: it looks like this block can be moved into a method and then reuse this method in other places where we use the same code block

iguannago (dismissed):

src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CFTTaskDatabaseServiceTest.java
```diff
@@ -53,20 +55,21 @@ void should_succeed_and_find_a_task_by_id() {
```
BenNzewi, line 55: **cftTaskDatabaseService.findByIdAndObtainPessimisticWriteLock(taskResource.getTaskId());**
If the task is locked and and another resource tries to obtain the lock on that task, is that handled elegantly and should  we include a test for that edge case (as per AC2)

BenNzewi (commented):
For "completion and cancellation" happy path FT's, should we update and verify tests to show that after either action and once camunda process is successful,  that the entries CFTTaskState is set. (Should be as it didn't roll back)

iguannago (approved):

PR #310, opened 2021-08-05
Task query access control - Grant type Basic

indvamshi:

https://tools.hmcts.net/jira/browse/RWA-673

### Change description ###

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/CftQueryService.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public List<TaskResource> getAllTasks(
+        SearchTaskRequest searchTaskRequest,
+        AccessControlResponse accessControlResponse,
+         List<PermissionTypes> permissionsRequired
+    ) {
+        final Specification<TaskResource> taskResourceSpecification = TaskResourceSpecification
+            .getTasks(searchTaskRequest, accessControlResponse, permissionsRequired);
+
+        Pageable page = PageRequest.of(0, 10, Sort.Direction.ASC, "taskId");
```
raghera, line 34: So should this not take the page and size params from the incoming request also?

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/TaskResourceSpecification.java
```diff
@@ -0,0 +1,161 @@
...
+            .and(searchByClassification(accessControlResponse)
+            .and(searchByRegion()
+            .and(searchByState(searchTaskRequest)
+            .and(searchByLocation(searchTaskRequest)
+            .and(searchByJurisdiction(searchTaskRequest)
+            .and(searchByCaseId(searchTaskRequest)
+            .and(searchByUser(searchTaskRequest))))))));
+    }
+
+    public static Specification<TaskResource> searchByRoles(List<PermissionTypes> permissionsRequired,
```
raghera, line 47: do these have to be static?

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/CftQueryService.java
indvamshi: Yes, it was hardcoded to test the pagination. But will change with RWA-591 ticket.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/TaskResourceSpecification.java
```diff
@@ -0,0 +1,161 @@
...
+    ) {
+
+        return searchByRoles(permissionsRequired, accessControlResponse)
+            .and(searchByClassification(accessControlResponse)
+            .and(searchByRegion()
+            .and(searchByState(searchTaskRequest)
+            .and(searchByLocation(searchTaskRequest)
+            .and(searchByJurisdiction(searchTaskRequest)
+            .and(searchByCaseId(searchTaskRequest)
+            .and(searchByUser(searchTaskRequest))))))));
```
raghera, line 44: Maybe I don't understand exactly how this works but just wanted to understand.  If we have multiple role assignments in the accessControlResponse then I guess we want to `OR` at the top level?  Is that what we are doing here?

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/CftQueryService.java
raghera: Just leave it out as a rule if it's in another ticket.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/TaskResourceSpecification.java
indvamshi: RoleAssignments condition can be placed anywhere in where clause and satisfy with multiple role assignment OR condition, for eg: based on above filters the hibernates generates query 

    where
        (
            taskresour0_.assignee in (
                ?
            )
        ) 
        and 1=1 
        and (
            taskresour0_.jurisdiction in (
                ?
            )
        ) 
        and (
            taskresour0_.location in (
                ?
            )
        ) 
        and (
            taskresour0_.state in (
                ?
            )
        ) 
        and taskresour0_.region=? 
        and (
            taskresour0_.security_classification in (
                ? , ? , ?
            )
        ) 
        **and (
            taskrolere1_.role_name=? 
            or taskrolere1_.role_name=?
        )** 
        and taskrolere1_.read=true 
    order by
        taskresour0_.task_id asc limit ?

indvamshi: check this condition in the above query 
        and (
            taskrolere1_.role_name=? 
            or taskrolere1_.role_name=?
        )

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/TaskResourceSpecification.java
indvamshi: The reason for using static is, The Specification interface itself has static methods like and, or, where..etc. So followed the same way of implementing filters. Moreover, the filters does not hold any state and don't see a benefit of creating instances to access those methods.

raghera: Ok since these sounds more like a helper class to create a specification with no instance variables, sounds like static methods are ok.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/TaskResourceSpecification.java
```diff
@@ -0,0 +1,200 @@
...
+    ) {
+
+        return searchByRoles(permissionsRequired, accessControlResponse)
+            .and(searchByState(searchTaskRequest)
+                .and(searchByLocation(searchTaskRequest)
+                    .and(searchByJurisdiction(searchTaskRequest)
+                        .and(searchByCaseId(searchTaskRequest)
+                            .and(searchByUser(searchTaskRequest)
+                                .and(searchByTaskType(searchTaskRequest)
+                                    .and(searchByTaskId(searchTaskRequest))))))));
```
raghera, line 54: I think the main search endpoint does not allow searching by `taskId` or `taskType`.  So I assume the intention then is to use this method for the search-for-completable endpoint and GET task/{task-id} endpoint?  Then are we going to put the search parameters into the `SearchTaskRequest`?  That may become quite confusing as we are mixing the params from different endpoints.

raghera (commented):
I've left some comments which you may be addressing already so take a look.  Happy to talk about them if that helps too.

Also I did not see any tests in here so is this more like a `Draft` PR for now?

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/CftQueryService.java
```diff
@@ -0,0 +1,81 @@
...
+        AccessControlResponse accessControlResponse,
+         List<PermissionTypes> permissionsRequired
+    ) {
+        final Specification<TaskResource> taskResourceSpecification = TaskResourceSpecification
+            .getTasks(searchTaskRequest, accessControlResponse, permissionsRequired);
+
+        final List<SortingParameter> sortingParameters = searchTaskRequest.getSortingParameters();
+
+        Sort sort;
+        if (sortingParameters == null || sortingParameters.isEmpty()) {
```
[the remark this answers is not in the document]
indvamshi: > Is the `null` check necessary? Maybe it's always an empty list?

The list is not initialised. so have to check for null.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/TaskResourceSpecification.java
indvamshi: It does allow . The SearchParameterKey has following enum values which user can search on

    LOCATION("location"),
    USER("user"),
    JURISDICTION("jurisdiction"),
    STATE("state"),
    TASK_ID("taskId"),
    TASK_TYPE("taskType"),
    CASE_ID("caseId");

PR #312, opened 2021-08-09
skip the permission check for the WA testing jurisdiction

iguannago:
https://tools.hmcts.net/jira/browse/RWA-691


iguannago:
it's not needed.

PR #340, opened 2021-09-21
New task query functionality plugin

indvamshi:

https://tools.hmcts.net/jira/browse/RWA-744

### Change description ###

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/CftQueryService.java
```diff
@@ -23,7 +26,7 @@ public CftQueryService(TaskResourceRepository taskResourceRepository) {
         this.taskResourceRepository = taskResourceRepository;
     }
 
-    public List<TaskResource> getAllTasks(
+    public GetTasksResponse<Task> getAllTasks(
```
mkayad, line 29: wouldn't be better to keep List<TaskResource> at this level and not use the GetTasksResponse which is the upper level?

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/CftQueryService.java
```diff
@@ -36,9 +39,29 @@ public CftQueryService(TaskResourceRepository taskResourceRepository) {
...
-        final Page<TaskResource> taskResources = taskResourceRepository.findAll(taskResourceSpecification, page);
+        final Page<TaskResource> pages = taskResourceRepository.findAll(taskResourceSpecification, page);
 
-        return taskResources.toList();
+        final List<TaskResource> taskResources = pages.toList();
+
+        return mapToTask(taskResources, pages.getTotalElements());
+    }
+
+    private GetTasksResponse<Task> mapToTask(List<TaskResource> taskResources, long totalNumberOfTasks) {
```
mkayad, line 49: Same thing here, why not keep the standard list and wrap the answer in GetTasksResponse in the controller

mkayad (changes requested):

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/CftQueryService.java
indvamshi: following the standards , making them consistent across the apis.

mkayad: Which standards are you referring to? i think class dependency should be downward not upward, it should not depend on class which are in the controllers package. We ask other devs if you want.

indvamshi: I got it. will change.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/CftQueryService.java
indvamshi: I got it. will change.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/CftQueryService.java
indvamshi: The reason being the totalNumber of records cannot be added separately in controller. For MVP , to get total number or records extra call made to get total number of records but in jpa case its overkill.

mkayad (dismissed):
LGTM

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/CftQueryService.java
```diff
@@ -36,9 +39,29 @@ public CftQueryService(TaskResourceRepository taskResourceRepository) {
...
+        final List<Task> tasks = taskResources.stream().map(taskResource ->
+            new Task(taskResource.getTaskId(), taskResource.getTaskName(), taskResource.getTaskType(),
+                taskResource.getState().getValue(), taskResource.getTaskSystem().getValue(),
+                taskResource.getSecurityClassification().getSecurityClassification(),
+                taskResource.getTitle(),
+                taskResource.getCreated() == null ? null : taskResource.getCreated().toZonedDateTime(),
```
BenNzewi, line 55: can we add unit tests to cover null scenarios for these 2 lines?

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/RoleAssignmentFilter.java
```diff
@@ -163,19 +173,21 @@ private static Predicate mapSecurityClassification(
                     SecurityClassification.PUBLIC
                 )
             );
-        } else {
+        } else if (classification.equals(Classification.PUBLIC)) {
```
BenNzewi, line 176: unit test for when !equal?

indvamshi: I have to move that else part, classification cannot be other than private, public and restricted.

PR #360, opened 2021-10-06
Feature/rwa 737 permission dmn

iguannago:
https://tools.hmcts.net/jira/browse/RWA-814


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/entities/NoteResource.java
```diff
@@ -30,7 +31,11 @@ private NoteResource() {
     }
 
     @JsonCreator
-    public NoteResource(String code, String noteType, String user, OffsetDateTime created, String content) {
+    public NoteResource(@JsonProperty("code") String code,
+                        @JsonProperty("noteType") String noteType,
+                        @JsonProperty("user") String user,
+                        @JsonProperty("created") OffsetDateTime created,
+                        @JsonProperty("content") String content) {
```
alexbaeza, line 38: Let's try to avoid the JsonProperty especially in this case it makes it camel case our APIs should only be using snake case

PR #440, opened 2021-11-22
Rwa 927 new

indvamshi:

https://tools.hmcts.net/jira/browse/RWA-927

### Change description ###

```
[ ] Yes
[ x] No
```


dharmendrak (dismissed):

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/CftQueryService.java
```diff
@@ -144,6 +144,7 @@ public CftQueryService(CamundaService camundaService, CFTTaskMapper cftTaskMappe
             .filter(result -> result.containsKey(TASK_TYPE.value()))
             .map(result -> camundaService.getVariableValue(result.get(TASK_TYPE.value()), String.class))
             .filter(Objects::nonNull)
+            .filter(val -> val.trim().length() != 0)
```
alexbaeza, line 147: I believe the getVariableValue() attempts to read the value into an optional and it can either return a variable value or a null. 

Not sure if we need to do this length check. was there a particular case e.g a value was empty ?

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/RoleAssignmentFilter.java
```diff
@@ -204,29 +209,38 @@ private static Predicate mapSecurityClassification(
...
-                    SecurityClassification.PUBLIC
-                )
-            );
-        } else {
-            return builder.in(root.get(SECURITY_CLASSIFICATION_COLUMN)).value(
-                List.of(
-                    SecurityClassification.PUBLIC
-                )
-            );
+        if (classification != null) {
```
alexbaeza, line 212: if the classification attribute is not provided should this fall under the default clause and search for everything PUBLIC/PRIVATE/RESTRICTED ?

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/RoleAssignmentFilter.java
```diff
@@ -204,29 +209,38 @@ private static Predicate mapSecurityClassification(
...
+                    );
+                default:
+                    return builder.in(root.get(SECURITY_CLASSIFICATION_COLUMN)).value(
+                        emptyList()
+                    );
+            }
         }
+        return builder.in(root.get(SECURITY_CLASSIFICATION_COLUMN)).value(
+            emptyList()
+        );
```
alexbaeza, line 243: could avoid this by using the default clause

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/TaskQuerySpecification.java
```diff
@@ -1,63 +1,167 @@
...
     }
 
     public static Specification<TaskResource> searchByUser(List<String> users) {
-        return (root, query, builder) -> builder.in(root.get(ASSIGNEE))
-            .value(users);
+        if (!CollectionUtils.isEmpty(users)) {
+            return (root, query, builder) -> builder.in(root.get(ASSIGNEE))
+                .value(users);
+        }
+        return (root, query, builder) -> builder.conjunction();
```
alexbaeza, line 75: personal preference would be to use it as a safe guard: what I mean:

instead of negating the CollectionUtils.isEmpty(users)

```suggestion
        if (CollectionUtils.isEmpty(users)) {
          return (root, query, builder) -> builder.conjunction();
        }
       return (root, query, builder) -> builder.in(root.get(ASSIGNEE))
                .value(users);
```

this way the logic is always default

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/TaskQuerySpecification.java
```diff
@@ -1,63 +1,167 @@
...
+                .map(value -> CFTTaskState.valueOf(value.toUpperCase(Locale.ROOT))).collect(Collectors.toList());
+
+            return searchByState(cftTaskStates);
+        }
+
+        return (root, query, builder) -> builder.conjunction();
+    }
+
+    public static Specification<TaskResource> extractJurisdiction(SearchTaskRequest searchTaskRequest) {
+        final EnumMap<SearchParameterKey, SearchParameter> keyMap = asEnumMap(searchTaskRequest);
```
alexbaeza, line 113: a lot of these extractors duplicate this line since we pass in a searchTaskRequest

What this means:
Each time that we call one of these extractors we are looping through the entire searchtaskRequest object in order to map to the Enum map.

I think it would be cleaner to loop through it only once and build the map and then pass it onto these extractors.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/exceptions/ServerErrorException.java
```diff
@@ -10,4 +10,10 @@ public ServerErrorException(
     ) {
         super(message, cause);
     }
+
+    public ServerErrorException(
+        String message
+    ) {
+        super(message);
+    }
```
alexbaeza, line 18: I could not find where this is used but in any case I believe this is a v1 exception we should try to use the v2 ones which have a application+problem json

alexbaeza (changes requested):
Some suggestions

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/CftQueryService.java
indvamshi: I was expecting that question :). Actually, pitest mutated with "" as response from DMN since we are not handling I had to filter the dmn response based on the "" and return empty results.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/RoleAssignmentFilter.java
indvamshi: TBH, am not sure what the business requirement is if its not provided. We have not tested though so far.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/exceptions/ServerErrorException.java
indvamshi: This is not required will remove it.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/TaskQuerySpecification.java
indvamshi: makes sense, will make changes.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/TaskQuerySpecification.java
indvamshi: With PMD check few times I see PMD warns to use negated condition first.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/CftQueryService.java
alexbaeza: Interesting I did not know about that in that case should we filter it once only with something like `StringUtils.hasText()` ?

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/CftQueryService.java
```diff
@@ -144,6 +145,7 @@ public CftQueryService(CamundaService camundaService, CFTTaskMapper cftTaskMappe
             .filter(result -> result.containsKey(TASK_TYPE.value()))
             .map(result -> camundaService.getVariableValue(result.get(TASK_TYPE.value()), String.class))
             .filter(Objects::nonNull)
+            .filter(StringUtils::hasText)
```
alexbaeza, line 148: can filter only once since hasText check for null :) 
```suggestion
            .filter(StringUtils::hasText)
```

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/TaskResourceSpecification.java
```diff
@@ -63,80 +67,41 @@ private TaskResourceSpecification() {
...
+        SearchParameter userParam = keyMap.get(USER);
+        SearchParameter workTypeParam = keyMap.get(WORK_TYPE);
+
+        List<CFTTaskState> cftTaskStates = new ArrayList<>();
+        if (stateParam != null) {
+            final List<String> values = stateParam.getValues();
+            if (!values.isEmpty()) {
+                cftTaskStates = values.stream()
+                    .filter(Objects::nonNull)
+                    .filter(StringUtils::hasText)
```
alexbaeza, line 93:
```suggestion
                    .filter(StringUtils::hasText)
```

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/CftQueryService.java
indvamshi: forgot to remove ..argh....doing it now.

alexbaeza (approved):

PR #446, opened 2021-11-24
RWA-948: Adds deleted terminate reason to mark tasks as TERMINATED 

alexbaeza:

https://tools.hmcts.net/jira/browse/RWA-948

### Change description ###

When a task is cancelled via the DMN the wa-case-event handler will create a cancellation message to be sent to the cancellation sub-process.

When this cancellation is received camunda now stores the cancelled task with a cancellation reason of 'deleted'.

As a side effect, this prevents R2 termination of tasks that have been cancelled automatically. Since the job never calls the terminate endpoint to terminate these tasks in the cft task db. 

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskManagementService.java
+++ b/src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskManagementService.java
@@ -637,6 +637,14 @@ public class TaskManagementService {
                 //Commit transaction
                 cftTaskDatabaseService.saveTask(task);
                 break;
+            case DELETED:
+                //Update cft task
+                task.setState(CFTTaskState.TERMINATED);
+                //Perform Camunda updates
+                camundaService.deleteCftTaskState(taskId);
+                //Commit transaction
+                cftTaskDatabaseService.saveTask(task);
+                break;
             default:
                 throw new IllegalStateException("Unexpected value: " + terminateInfo.getTerminateReason());
         }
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskManagementService.java
```diff
@@ -637,6 +637,14 @@ public void terminateTask(String taskId, TerminateInfo terminateInfo) {
...
                 cftTaskDatabaseService.saveTask(task);
                 break;
+            case DELETED:
+                //Update cft task
+                task.setState(CFTTaskState.TERMINATED);
+                //Perform Camunda updates
+                camundaService.deleteCftTaskState(taskId);
+                //Commit transaction
+                cftTaskDatabaseService.saveTask(task);
+                break;
```
raghera, line 647: Maybe we need a fallback rule?  e.g. If it's not a known reason maybe that should be set to Terminated.

raghera: We may also want to check the states we are setting: https://tools.hmcts.net/confluence/pages/viewpage.action?pageId=1518684941#HLDTaskRepositoryv1.1-2.5.2CFTTaskState

alexbaeza: This could be a clean solution to default to terminated, happy to implement if needed

alexbaeza: Discussion was taken offline and a new ticket was raised to attempt to resolve and cater for future error scenarios, 
Thanks @raghera

indvamshi (approved):

PR #534, opened 2022-03-03
Adding index on task_roles task_id foreign key

raghera:
Index on foreign key for performance improvement.


skandala-cgi:
Closing thus PR as it is a draft PR and raised long time back

PR #569, opened 2022-04-01
Change IN to Equal in CFT queries  for single elements

dharmendrak:
Jira: https://tools.hmcts.net/jira/browse/RWA-1264


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/query/TaskQuerySpecification.java
```diff
@@ -29,25 +30,28 @@ private TaskQuerySpecification() {
     public static Specification<TaskResource> searchByState(List<CFTTaskState> cftTaskStates) {
         if (isEmpty(cftTaskStates)) {
             return (root, query, builder) -> builder.conjunction();
+        } else if (cftTaskStates.size() == SINGLE_ELEMENT) {
+            return (root, query, builder) -> builder.equal(root.get(STATE), cftTaskStates.get(0));
         }
-        return (root, query, builder) -> builder.in(root.get(STATE))
-            .value(cftTaskStates);
```
raghera, line 34: Can we add some logging so we can see how many times we hit the IN vs EQUALS clauses?

raghera (approved):

PR #607, opened 2022-05-10
RWA-1154 - Application insights upgrade to 3.2.11

AndreiTrushkov:
https://tools.hmcts.net/jira/browse/RWA-1154

### Change description ###
application insights upgrade to 3.2.11
various dependencies upgrade

```
[ ] Yes
[ x] No
```


lib/applicationinsights.json
```diff
@@ -0,0 +1,6 @@
+{
+  "connectionString": "${file:/mnt/secrets/wa/app-insights-connection-string}",
+  "role": {
+    "name": "WA Task Management"
```
raghera, line 4: Assume this role name is not standardised/used anywhere else and can be any string?

raghera (approved):
Looks ok to me.  Just a small query but approved.

lib/applicationinsights.json
AndreiTrushkov: https://docs.microsoft.com/en-us/azure/azure-monitor/app/java-standalone-config#cloud-role-name  used as label for component in application map, i've used app name for it

PR #713, opened 2022-08-16
RWA-1712

AndreiTrushkov:
### Change description ###
java 17 upgrade

```
[ ] Yes
[x] No
```


config/pmd/ruleset.xml
```diff
@@ -9,6 +9,8 @@
   <rule ref="category/java/bestpractices.xml">
     <exclude name="GuardLogStatement"/>
     <exclude name="JUnitTestContainsTooManyAsserts"/>
+    <exclude name="LiteralsFirstInComparisons"/>
+    <exclude name="UnusedAssignment"/>
```
raghera, line 13: I can't see the errors but can we not fix some of these exclusions?  Unused assignments seems like something easy to get rid of?

config/pmd/ruleset.xml
```diff
@@ -64,6 +68,7 @@
   </rule>
   <rule ref="category/java/errorprone.xml">
     <exclude name="BeanMembersShouldSerialize"/>
+    <exclude name="ReturnEmptyCollectionRatherThanNull"/>
```
raghera, line 71: Another one that maybe we could fix?

config/pmd/ruleset.xml
```diff
@@ -41,6 +43,8 @@
     <exclude name="UseUtilityClass"/>
     <exclude name="LoosePackageCoupling"/>
     <exclude name="DataClass"/>
+    <exclude name="CognitiveComplexity"/>
+    <exclude name="CyclomaticComplexity"/>
```
raghera, line 47: Do you have an example of code that got flagged by this?
Again in general would be good if we don't turn these off centrally.

config/pmd/ruleset.xml
abhiahirv1: Yes, I rid of it

config/pmd/ruleset.xml
abhiahirv1: Sure, it is removed

config/pmd/ruleset.xml
abhiahirv1: I removed it

abhiahirv1 (commented):
Removing extra dependencies

raghera (approved):

PR #778, opened 2022-10-05
RWA-1852 SAR - Next step is not working

mkayad:
[RWA-1852](https://tools.hmcts.net/jira/browse/RWA-1852)

### Change description ###

see ticket 

```
[ ] Yes
[ X] No
```


src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/watasks/controllers/GetTaskByIdControllerTest.java
```diff
@@ -363,4 +363,127 @@ public void should_replace_additional_properties_in_configuration_dmn_and_return
         common.cleanUpTask(taskId);
     }
 
+    @Test
+    public void should_return_a_200_with_task_id_in_description_when_creating_standalone_task() {
+        String taskType = "reviewSpecificAccessRequestLegalOps";
+        String taskName = "review specific access request legal ops";
+        TestVariables taskVariables = common.setupWAStandaloneTaskAndRetrieveIds(
```
raghera, line 370: Is this just checking in Camunda?

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/taskmanagementservicetests/NewInitiateTaskTest.java
```diff
@@ -149,6 +149,7 @@ void given_initiateTask_then_task_is_saved() {
         OffsetDateTime dueDate = OffsetDateTime.now();
         String formattedDueDate = CAMUNDA_DATA_TIME_FORMATTER.format(dueDate);
         initiateTaskRequest.getTaskAttributes().put(DUE_DATE.value(), formattedDueDate);
+        initiateTaskRequest.getTaskAttributes().put("taskId", taskId);
```
raghera, line 152: I think the Task Monitor has to make sure that the taskId value is populated correctly.  Is this done already?

src/testUtils/java/uk/gov/hmcts/reform/wataskmanagementapi/config/GivensBuilder.java
```diff
@@ -73,12 +82,20 @@ public GivensBuilder iCreateATaskWithCustomVariables(Map<String, CamundaValue<?>
...
                                                 String taskType, String taskName) {
         Map<String, CamundaValue<?>> processVariables = createDefaultTaskVariables(caseId, jurisdiction, caseType,
-                                                                                   taskType, taskName, Map.of()
+            taskType, taskName, Map.of()
         );
         createTask(processVariables);
         return this;
     }
 
+    public GivensBuilder iCreateAStandaloneTaskWithCaseId(String caseId, String jurisdiction, String caseType,
```
raghera, line 91: I think the naming can be better here.  This is a common file that everyone will want to use and so it needs to be absolutely obvious what it is doing.
Looks to me like it is triggering the Standalone BPMN process via the Workflow api so it should be named like that.  Eg. iSendAMessageToWorkflowApi.  

Better would be to refactor the other methods too so you can make it all more generic.  But due to the urgency of this PR it's better that is discussed later.

raghera (commented):
In general it looks ok.  
Maybe I missed but is there a functional test that creates a Specific request Task and then the review Task?  Think I only saw tests that create a Task using the Workflow API and Case Events.

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/watasks/controllers/GetTaskByIdControllerTest.java
```diff
@@ -363,4 +363,127 @@ public void should_replace_additional_properties_in_configuration_dmn_and_return
         common.cleanUpTask(taskId);
     }
 
+    @Test
+    public void should_return_a_200_with_task_id_in_description_when_creating_standalone_task() {
```
rasikasri, line 367: I think both these tests more suitable in PostTaskInitiateByIdControllerTest. Am I right thinking that discription is created in task initiation process and the get task returning whatever the task info in cft_task db? If that's the case, there is nothing new in get operation here than the other tests in the same class. Also, it is better if you can test the discription in any existing test rather than adding a new one (only if you can).

mkayad: I think you got a point, we are touching the initiation endpoint, so will move the test PostTaskInitiateByIdControllerTest

src/testUtils/java/uk/gov/hmcts/reform/wataskmanagementapi/config/GivensBuilder.java
mkayad: will rename these methods and let you know

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/watasks/controllers/GetTaskByIdControllerTest.java
mkayad: no it is not, it is checking CFT db, in Camunda, the task Id is the same as the task type

rasikasri (dismissed):

dharmendrak (approved):

PR #792, opened 2022-10-17
RWA-1519 remove R1 elements

vgulapally:
https://tools.hmcts.net/jira/browse/RWA-1519

### Change description ###
Remove R1 relements and refactor code

```
[ ] Yes
[x] No
```


raghera (approved):
Walked through with Vamshi on call and looks good.
We noticed that there is still the old R1 Task Configuration code which requires removal but we think this can be done in another ticket.  This one is already a large change.

src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/GetTaskByIdControllerTest.java
```diff
@@ -130,35 +118,10 @@ void setUp() {
         );
     }
 
-    @Test
-    void should_return_a_500_when_id_invalid() throws Exception {
```
BenNzewi, line 134: can we please check invalid id check is covered in the R2 code? I.e. against CFTTaskDb

src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskCompleteByIdControllerFailureTest.java
```diff
@@ -506,59 +541,56 @@ void beforeEach() {
                 .thenReturn(false);
         }
 
-        @Test
-        void should_return_500_with_application_problem_response_when_task_update_call_fails() throws Exception {
```
BenNzewi, line 510: Another one? This covers when we fail to complete and fails to update the task. Are covered on cftTaskDb side?

vgulapally: done

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskManagementServiceTest.java
```diff
@@ -2606,287 +1525,6 @@ void should_throw_task_not_found_exception_when_task_resource_not_found_and_feat
         }
     }
 
-    @Nested
-    @DisplayName("searchWithCriteria()")
```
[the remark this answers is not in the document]
vgulapally: @DisplayName("searchWithCriteria()") in task management service is completed removed as it calls camunda api, instead now its cftqueryservice searchwithcriteria, already tests covered for thats in CFTqueryServiceTests

Since these method is completely removed from TaskManagementService, corresponding tests were also removed
so for R2 calls are only made from CFTQueryService not from TaskManagementService

chrisdavecm (approved):

PR #818, opened 2022-11-02
RWA-1658 task action attributes for complete endpoint

vgulapally:
https://tools.hmcts.net/jira/browse/RWA-1658

### Change description ###
added task action attributes for Complete endpoint

```
[ ] Yes
[x] No
```


src/testUtils/java/uk/gov/hmcts/reform/wataskmanagementapi/utils/Assertions.java
```diff
@@ -65,6 +71,57 @@ public void taskStateWasUpdatedInDatabase(String taskId, String value, Headers a
...
+            .statusCode(HttpStatus.OK.value())
+            .and().contentType(MediaType.APPLICATION_JSON_VALUE).log();
+
+        fieldValueMap.entrySet().forEach(
+            entry -> result.then().assertThat()
+                .body(entry.getKey(), entry.getValue()).log()
+        );
+    }
+
+    public void taskActionAttributesUpdatedInDatabase(String taskId, String assigneeId,
```
raghera, line 93: Should this be in this class called `Assertions`?  Doesn't assert anything.  Seems to prepare a list of attributes to assert later which are passed in as individual params.

src/testUtils/java/uk/gov/hmcts/reform/wataskmanagementapi/utils/Assertions.java
```diff
@@ -65,6 +71,57 @@ public void taskStateWasUpdatedInDatabase(String taskId, String value, Headers a
             .log();
     }
 
+    public void taskValuesWasUpdatedInDatabase(String taskId, Map<String, Matcher<?>> fieldValueMap,
```
raghera, line 74: Maybe rename to something like `taskAttributesVerifier`?

src/testUtils/java/uk/gov/hmcts/reform/wataskmanagementapi/utils/Assertions.java
```diff
@@ -65,6 +71,57 @@ public void taskStateWasUpdatedInDatabase(String taskId, String value, Headers a
...
+    public void nullTaskActionAttributesUpdatedInDatabase(String taskId, String taskState,
+                                                          TestAuthenticationCredentials taskRetrieveCredentials) {
+
+        Map<String, Matcher<?>> taskValueMap = Map.of(
+            "task.id", equalTo(taskId),
+            "task.task_state", CoreMatchers.is(taskState),
+            "task.assignee", nullValue(),
+            "task.last_updated_timestamp", nullValue(),
+            "task.last_updated_user", nullValue(),
+            "task.last_updated_action", nullValue()
```
raghera, line 118: Just to say when we switch this on we don't want the last_updated values to ever be `null` in the DB.  This makes things a lot simpler for CTSC MI.

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/watasks/controllers/PostTaskCompleteByIdControllerTest.java
```diff
@@ -578,6 +633,20 @@ public void user_should_not_complete_task_when_grant_type_specific_and_permissio
             .body("status", equalTo(403))
             .body("detail", equalTo(ROLE_ASSIGNMENT_VERIFICATION_DETAIL_REQUEST_FAILED));
 
+        String serviceToken = caseworkerCredentials.getHeaders().getValue(AUTHORIZATION);
+        UserInfo userInfo = authorizationProvider.getUserInfo(serviceToken);
+        //This will be changed when assign endpoint story RWA-1582 is played
+        Map<String, Matcher<?>> taskValueMap = Map.of(
```
raghera, line 639: Should we consider the deployment order if you're dependent on another ticket?  We don't want to have `null` values being set in the DB.  
When this feature is live every update to the DB will have an Action added and so there should be values populated here already.

Maybe a better solution here would be to check the state of the values before you run the test case and then verify that the values have not changed?  This way it is independent of other changes.

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/watasks/controllers/PostTaskCompleteByIdControllerTest.java
```diff
@@ -606,6 +675,20 @@ public void user_should_not_complete_task_when_grant_type_challenged_and_permiss
...
```
raghera, line 681: As above.

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/watasks/controllers/PostTaskCompleteByIdControllerTest.java
```diff
@@ -625,15 +708,20 @@ private void assignTask(TestVariables taskVariables) {
...
 
         assertions.taskVariableWasUpdated(taskVariables.getProcessInstanceId(), "taskState", "assigned");
-        assertions.taskStateWasUpdatedInDatabase(
-            taskVariables.getTaskId(), "assigned", caseworkerCredentials.getHeaders()
-        );
-        assertions.taskFieldWasUpdatedInDatabase(
-            taskVariables.getTaskId(), "assignee", assigneeId, caseworkerCredentials.getHeaders()
+
+        //This need to be changed when assign endpoint story RWA-1582 is played
+        Map<String, Matcher<?>> taskValueMap = Map.of(
```
raghera, line 713: again

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/watasks/controllers/PostTaskUnclaimByIdControllerTest.java
```diff
@@ -64,16 +74,27 @@ public void should_return_a_204_when_unclaiming_a_task_by_id_gp_flag_on() {
 
         assertions
             .taskVariableWasUpdated(taskVariables.getProcessInstanceId(), "taskState", "unassigned");
-        assertions.taskStateWasUpdatedInDatabase(taskId, "unassigned", caseworkerCredentials.getHeaders());
-        assertions.taskFieldWasUpdatedInDatabase(taskId, "assignee", null, caseworkerCredentials.getHeaders());
+
+        String serviceToken = unassignUser.getHeaders().getValue(AUTHORIZATION);
+        UserInfo userInfo = authorizationProvider.getUserInfo(serviceToken);
+
+        Map<String, Matcher<?>> valueMap = Map.of(
```
raghera, line 81: as above

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CFTTaskMapperTest.java
```diff
@@ -1145,6 +1148,9 @@ void should_map_task_resource_to_task_with_permissions_empty() {
         assertTrue(task.getPermissions().getValues().isEmpty());
         assertNotNull(task.getReconfigureRequestTime());
         assertNotNull(task.getLastReconfigurationTime());
+        assertNotNull(task.getLastUpdatedTimestamp());
```
raghera, line 1151: These should check the actual values not just if they're not null.

raghera (changes requested):
Looks good but a few comments.  Take a look.

src/testUtils/java/uk/gov/hmcts/reform/wataskmanagementapi/utils/Assertions.java
vgulapally: true, as we are currently doing one endpoint at a time, after we address all the endpoints probably this is true that these attributes wont be null, but at this point of time when task is initially created these values are null hence asserting it as null, once this method is no longer used will remove.

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/watasks/controllers/PostTaskCompleteByIdControllerTest.java
vgulapally: true, currently when task is initially created they are null, since we have done only claim, unclaim, complete endpoints, only those actions are saved in the DB, but once all endpoints are addressed we can see the gaps which endpoints are not covered or which of the cases these attributes can be null, probably we can discuss in a call?

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CFTTaskMapperTest.java
vgulapally: done

src/testUtils/java/uk/gov/hmcts/reform/wataskmanagementapi/utils/Assertions.java
vgulapally: just for reusing across different functional tests, i have added it here, currently its been used only by postCompletion functional test, later this method will be used in all other functional tests, can be used by others in future

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/watasks/controllers/PostTaskCompleteByIdControllerTest.java
vgulapally: I think we will connect and discuss regarding the best approach to deal with these?

vgulapally: done

src/testUtils/java/uk/gov/hmcts/reform/wataskmanagementapi/utils/Assertions.java
vgulapally: done

src/testUtils/java/uk/gov/hmcts/reform/wataskmanagementapi/utils/Assertions.java
vgulapally: this will be addressed in the configuration and auto-assignment story as discussed today

raghera (dismissed):

rasikasri (dismissed):

rasikasri (dismissed):

rasikasri (dismissed):

rasikasri (approved):

PR #857, opened 2022-11-24
RWA-2037 change the order of repos so that dependencies can be resolved

mcareyma:
https://tools.hmcts.net/jira/browse/RWA-2037

### Change description ###
Change the order of repositories so that dependencies that cannot be resolved in jitpak.io can be resolved in mavenCentral.

```
[ ] Yes
[X] No
```


mrmsali (approved):
Looks good to me.

PR #885, opened 2022-12-09
RWA-2093 update the audit.json for security scan problem

DennisHoV1:
https://tools.hmcts.net/jira/browse/RWA-2093

### Change description ###
-update the audit.json for security scan problem

```
[ ] Yes
[X] No
```


mcareyma (approved):
This is the answer though it seems a bit bonkers as it is just ignoring the issue

PR #886, opened 2022-12-09
[DONOT MERGE] Data migration for granular permissions and making it backword compat…

dharmendrak:

https://tools.hmcts.net/jira/browse/RWA-2068

### Change description ###

Data migration for granular permissions and making it backword compatibility

```
[ ] Yes
[ ] No
```


src/main/resources/db/migration/V1.0.23__update_permissions_from_task_roles_for_backward_compatibility.sql
```diff
@@ -0,0 +1,5 @@
+UPDATE task_roles SET unassign=true, assign=true, complete=true WHERE manage=true;
+
+UPDATE task_roles SET claim=true WHERE own=true;
+
+UPDATE task_roles SET claim=true WHERE execute=true;
```
raghera, line 5: Does not matter but could have combined last 2.

Also we only strictly need to run a migration on historic Tasks.  The _current_ ones (when we actually run it in prod) will have the new permissions already since we will have already deployed the new DMNs.

raghera: We may want to modify the script so we only touch the records that need to migrated.

PR #914, opened 2023-01-17
RWA-2076 - remove privileged-access-feature

Lee365:
https://tools.hmcts.net/jira/browse/RWA-2076

### Change description ###
RWA-2076 - remove privileged-access-feature

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/auth/restrict/ClientAccessControlService.java
```diff
@@ -16,53 +13,16 @@
...
-    /**
-     * Extracts client id from service authorization token and returns if client is whitelisted as privilegedServices.
-     * Note: This feature is sitting behind feature flag.
-     *
-     * @param serviceAuthToken      the service authorization token.
-     * @param accessControlResponse the access control response containing userId.
-     * @return whether a client has been whitelisted in config.privilegedAccessClients property.
-     */
-    @SuppressWarnings("PMD.DataflowAnomalyAnalysis")
-    public boolean hasPrivilegedAccess(String serviceAuthToken, AccessControlResponse accessControlResponse) {
```
raghera, line 45: Not sure if this is correct.
I think the requirement was to remove the feature flag rather than the functionality of checking that the client has permission to make this request.

So I would expect that the if/else logic on here based on a feature flag is removed but not the whole method.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/TaskActionsController.java
```diff
@@ -206,18 +206,11 @@ public ResponseEntity<Void> completeTask(@RequestHeader(AUTHORIZATION) String au
...
-                    completeTaskRequest.getCompletionOptions()
-                );
-            } else {
-                throw new GenericForbiddenException(GENERIC_FORBIDDEN_ERROR);
-            }
+            taskManagementService.completeTaskWithPrivilegeAndCompletionOptions(
+                taskId,
+                accessControlResponse,
+                completeTaskRequest.getCompletionOptions()
+            );
```
raghera, line 213: As per my point above, this will allow anyone to make a privileged request.  What we want is that _only_ ExUI can make these types of request.  We should continue to return a 403 when it's not.

raghera (changes requested):
See my comments on the code.  It most likely means that a lot of the tests removed will need to be reinstated too.

dharmendrak (approved):

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/auth/restrict/ClientAccessControlService.java
Lee365: Method added back. Only feature flag check removed.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/TaskActionsController.java
Lee365: Methods added back. Only feature flag check removed.

raghera (approved):
Looks good.

rasikasri (approved):

PR #926, opened 2023-02-02
Rwa 1835

chrisdavecm:
https://tools.hmcts.net/jira/browse/RWA-1835

### Change description ###

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/cft/entities/TaskHistoryResource.java
```diff
@@ -0,0 +1,68 @@
...
+import javax.persistence.Entity;
+import javax.persistence.Id;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+@ToString
+@Getter
+@EqualsAndHashCode(onlyExplicitlyIncluded = true)
+@Entity(name = "taskHistory")
+@SuppressWarnings({"PMD.TooManyFields"})
+public class TaskHistoryResource {
```
raghera, line 19: Not important but could this be a Java 17 Record?  There's some customisation required for equalandhashcode so may be more trouble than it's worth.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/response/TaskHistoryResponse.java
```diff
@@ -0,0 +1,14 @@
+package uk.gov.hmcts.reform.wataskmanagementapi.controllers.response;
+
+import lombok.Builder;
+import lombok.Data;
+import uk.gov.hmcts.reform.wataskmanagementapi.cft.entities.TaskHistoryResource;
+
+import java.util.List;
+
+@Builder(toBuilder = true)
+@Data
```
raghera, line 10: As above we don't need to use this any more with `Record`s in Java 17.  Don't need to fix here if easier.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/schedulers/LogicalReplicationCreatorScheduler.java
```diff
@@ -0,0 +1,24 @@
...
+public class LogicalReplicationCreatorScheduler {
+    private static final org.slf4j.Logger LOGGER = LoggerFactory.getLogger(LogicalReplicationCreatorScheduler.class);
+
+    @Autowired
+    private MIReportingService miReportingService;
+
+    @Scheduled(fixedDelay = 1000)
+    public void scheduled() {
+        LOGGER.debug("Postgresql logical replication scheduler executed");
+        miReportingService.logicalReplicationCheck();
```
raghera, line 22: Will this check every second even if the previous check has blocked for a few seconds?
I am just thinking about what happens in that case?  Could you get 2 attempts to create a publication?  Does that matter or does it just work?

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/MIReportingService.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    public List<TaskHistoryResource> findByTaskId(String taskId) {
+        return taskHistoryRepository.getByTaskId(taskId);
+    }
+
+    public void logicalReplicationCheck() {
+        LOGGER.debug("Postgresql logical replication check executed");
+        if (isReplicationSlotPresent()) {
+            if (!isPublicationPresent()) {
```
raghera, line 59: As the comment above.  Could you get multiple threads attempting to do this at the same time?  Will this have unwanted side effects?

raghera (approved):
Since this is an inactive version of the changes I have approved as long as the comments are considered in the active version later.

The comments are mainly questions so should not block this PR from moving ahead.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/schedulers/LogicalReplicationCreatorScheduler.java
chrisdavecm: Should be transactional

chrisdavecm: Have run it on perftest and no problems

rasikasri (approved):

PR #961, opened 2023-02-23
RWA-2212: Feature Flag - create a new feature flag

rasikasri:
### Change description ###

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskManagementService.java
```diff
@@ -775,6 +776,16 @@ public void terminateTask(String taskId, TerminateInfo terminateInfo) {
...
     }
 
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    public TaskResource updateTaskIndex(String taskId) {
+        TaskResource taskResource = cftTaskDatabaseService.findByIdOnly(taskId).orElse(null);
+        if (taskResource != null) {
+            taskResource.setIndexed(true);
+            cftTaskDatabaseService.saveTask(taskResource);
+        }
+        return taskResource;
```
raghera, line 786: Is there a potential race condition here?
If you have 2 threads both reading the same data with slightly different updates then what would happen?
Postgres has a default transaction isolation level of read_committed and we are saying here that these would be separate transactions.  So whoever wins the race would update and the 2nd thread will update it again in another transaction.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/repository/TaskResourceCustomRepositoryImpl.java
```diff
@@ -60,6 +61,7 @@ public List<String> searchTasksIds(int firstResult,
             PAGINATION_CLAUSE
         );
 
+        log.debug("Task search query [{}]", queryString);
```
raghera, line 64: might be useful to start this out at `info` level if no sensitive info.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskManagementService.java
```diff
@@ -775,6 +776,16 @@ public void terminateTask(String taskId, TerminateInfo terminateInfo) {
         }
     }
 
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    public TaskResource updateTaskIndex(String taskId) {
+        TaskResource taskResource = cftTaskDatabaseService.findByIdOnly(taskId).orElse(null);
+        if (taskResource != null) {
```
raghera, line 782: Minor point, but if you're getting an Optional back then why not just use it to check if the result is present?  Usually better to avoid `null` where possible since it is ambiguous.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/repository/TaskResourceCustomRepositoryImpl.java
rasikasri: Done

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskManagementService.java
rasikasri: Done

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskManagementService.java
rasikasri: changed the findTaskById method to findByIdAndObtainPessimisticWriteLock where it adds a lock before update the indexed flag. I am going to add a timeout to aquire a lock In RWA-2216 as I have done some changes in updateTaskIndex method there.

PR #969, opened 2023-02-28
RWA-2216: Update index actions

rasikasri:

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/request/enums/TaskOperationType.java
```diff
@@ -0,0 +1,23 @@
+package uk.gov.hmcts.reform.wataskmanagementapi.controllers.request.enums;
+
+public enum TaskOperationType {
+    MARK_TO_RECONFIGURE("mark_to_reconfigure", false),
+    EXECUTE_RECONFIGURE("execute_reconfigure", true);
```
raghera, line 5: Maybe add a comment here for devs who may add a new operation in future.  I guess this boolean is to determine whether the operation requires a re-indexing of signatures, which for new devs may be unclear since it is complex.

raghera: This is not a blocker for this PR but ... It would be nice to be able to abstract this complexity somehow as most new devs don't have to worry about it.  Is there a default that makes sense?  E.g. it's always true unless you specifically opt out?  Maybe that will just mean redundant re-indexing.  Or perhaps we could have some type of meaningful operation category (Task/Role update vs no Task / Role update) which means that the indexing happens or not.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskManagementService.java
```diff
@@ -777,15 +778,21 @@ public void terminateTask(String taskId, TerminateInfo terminateInfo) {
...
-            throw new TaskNotFoundException(TASK_NOT_FOUND_ERROR);
+            if (findTaskResponse.isPresent()) {
+                TaskResource taskResource = findTaskResponse.get();
+                taskResource.setIndexed(indexed);
+
+                cftTaskDatabaseService.saveTask(taskResource);
+            } else {
+                throw new TaskNotFoundException(TASK_NOT_FOUND_ERROR);
+            }
+        } catch (LockTimeoutException ex) {
```
raghera, line 794: I think RWA-2332 will address this, but we can possibly get other Sql exceptions here.

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CFTTaskDatabaseServiceTest.java
```diff
@@ -80,6 +80,20 @@ void should_find_by_id_and_obtain_pessimistic_write_lock() {
         assertEquals(someTaskResource, actualTaskResource.get());
     }
 
+    @Test
+    void should_find_by_id_and_wait_and_obtain_pessimistic_write_lock() {
```
raghera, line 84: There are probably more scenarios to test ... but again I think the next ticket ought to cover those.

rasikasri: added a test for exception scenario

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskManagementService.java
rasikasri: Changed to handle PersistenceException

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/request/enums/TaskOperationType.java
rasikasri: Added a comment to explain boolean flag. As you have said, we can add operation types in a different ticket.

PR #993, opened 2023-03-22
RWA-2398 Task Types endpoint not working as specified

mkayad:

[RWA-2398](https://tools.hmcts.net/jira/browse/RWA-2398]

### Change description ###

see ticket

```
[ ] Yes
[ X] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/TaskTypesControllerTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/TaskTypesControllerTest.java
@@ -26,6 +26,7 @@ import java.util.List;
 
 import static java.util.Collections.singletonList;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyBoolean;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.lenient;
 import static org.mockito.Mockito.when;
@@ -188,7 +189,8 @@ class TaskTypesControllerTest extends SpringBootIntegrationBaseTest {
         when(camundaServiceApi.getTaskTypesDmnTable(
             anyString(),
             anyString(),
-            anyString()
+            anyString(),
+            anyBoolean()
         )).thenThrow(FeignException.FeignServerException.class);
 
         mockMvc.perform(
@@ -221,7 +223,8 @@ class TaskTypesControllerTest extends SpringBootIntegrationBaseTest {
         when(camundaServiceApi.getTaskTypesDmnTable(
             anyString(),
             anyString(),
-            anyString()
+            anyString(),
+            anyBoolean()
         )).thenThrow(FeignException.ServiceUnavailable.class);
 
         mockMvc.perform(
--- a/src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/clients/CamundaServiceApi.java
+++ b/src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/clients/CamundaServiceApi.java
@@ -179,7 +179,8 @@ public interface CamundaServiceApi {
     List<TaskTypesDmnResponse> getTaskTypesDmnTable(
         @RequestHeader(SERVICE_AUTHORIZATION) String serviceAuthorisation,
         @RequestParam("tenantIdIn") String jurisdiction,
-        @RequestParam("name") String dmnName
+        @RequestParam("keyLike") String dmnKey,
+        @RequestParam("latestVersion") boolean latestVersion
     );
 
     @GetMapping(
--- a/src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/DmnEvaluationService.java
+++ b/src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/DmnEvaluationService.java
@@ -133,26 +133,31 @@ public class DmnEvaluationService {
         }
     }
 
-    private Set<TaskTypesDmnResponse> performRetrieveTaskTypesDmn(String jurisdiction, String dmnKeyLikeField) {
+    private Set<TaskTypesDmnResponse> performRetrieveTaskTypesDmn(String jurisdiction, String dmnDecisionKey) {
         try {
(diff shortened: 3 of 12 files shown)


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/clients/CamundaServiceApi.java
```diff
@@ -179,7 +179,8 @@ List<ConfigurationDmnEvaluationResponse> evaluateConfigurationDmnTable(
     List<TaskTypesDmnResponse> getTaskTypesDmnTable(
         @RequestHeader(SERVICE_AUTHORIZATION) String serviceAuthorisation,
         @RequestParam("tenantIdIn") String jurisdiction,
-        @RequestParam("name") String dmnName
+        @RequestParam("keyLike") String dmnKey,
+        @RequestParam("latestVersion") boolean latestVersion
```
raghera, line 183: 👍

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/DmnEvaluationService.java
```diff
@@ -130,25 +133,27 @@ private List<PermissionsDmnEvaluationResponse> performEvaluatePermissionsDmnActi
...
-    private Set<TaskTypesDmnResponse> performRetrieveTaskTypesDmn(String jurisdiction, String dmnNameField) {
+    private Set<TaskTypesDmnResponse> performRetrieveTaskTypesDmn(String jurisdiction, String dmnKeyLikeField) {
         try {
             List<TaskTypesDmnResponse> taskTypesDmnResponseList = camundaServiceApi.getTaskTypesDmnTable(
                 serviceAuthTokenGenerator.generate(),
                 jurisdiction.toLowerCase(Locale.ROOT),
-                dmnNameField
+                dmnKeyLikeField,
+                true
             );
```
raghera, line 143: Would be good to log the keys and ids you found for debugging in future?

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskTypesService.java
```diff
@@ -51,7 +51,7 @@ public GetTaskTypesResponse getTaskTypes(AccessControlResponse accessControlResp
 
         //get task-type-dmn(s) for jurisdiction
         Set<TaskTypesDmnResponse> taskTypesDmnResponse =
-            dmnEvaluationService.retrieveTaskTypesDmn(jurisdiction, DMN_NAME);
+            dmnEvaluationService.retrieveTaskTypesDmn(jurisdiction, DECISION_KEY_PREFIX);
```
raghera, line 54: Do we do any duplicate checking of the files we get back?  E.g. if 2 get 2 DMNs back and there is the same id in both and say they have different values?

src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/TaskTypesControllerTest.java
```diff
@@ -221,7 +223,8 @@ void should_return_503_with_application_problem_response_when_camunda_api_is_not
         when(camundaServiceApi.getTaskTypesDmnTable(
             anyString(),
             anyString(),
-            anyString()
+            anyString(),
+            anyBoolean()
```
raghera, line 227: Do we have DMN tests where multiple DMNs are returned?

raghera (commented):
Generally looks good.  Some questions.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/DmnEvaluationService.java
mkayad: will add a debug log

src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/TaskTypesControllerTest.java
mkayad: no we don't have it at the moment.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskTypesService.java
mkayad: no we don't do duplicate checking. unless the service have 2 files with the same id, the latestVersion parameter will filter out the old one.

src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/TaskTypesControllerTest.java
raghera: Again this can be a follow up ... I don't think any team uses it currently but it will happen in future.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskTypesService.java
mkayad: Replaced the list by a set to avoid duplicate files

raghera (approved):
So the downside of the Set is that it will lose the order of the Task types you receive unless it’s a SortedSet however, I think the implementation of this feature on the front end is not a dropdown but a type ahead textbox … so then the order should not matter.  So looks good.

src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/TaskTypesControllerTest.java
BenNzewi, line 227: One of the AC's is to handle the latest version of the dmn I.e. "about evaluating the latest version of each DMN and there is a flag called `latestVersion`". Can we add a test for this to show we always take the latest?

mkayad: As discussed, the latest version parameter should return the latest version. this managed within Camunda

PR #1004, opened 2023-03-27
Update Swagger publisher

timja:


.github/workflows/swagger.yml
```diff
@@ -7,32 +7,29 @@ on:
...
         run: |
           mkdir swagger-staging
             cd swagger-staging
           git init
-          git config user.email "jenkins-reform-iac@hmcts.net"
-          git config user.name "WA Task Management API"
-          git remote add upstream https://${{ secrets.SWAGGER_PUBLISHER_API_TOKEN }}@github.com/hmcts/reform-api-docs.git
+          git config user.email "github-platform-operations@HMCTS.NET"
+          git config user.name "HMCTS Platform Operations"
+          git remote add upstream https://apitoken:${{ secrets.SWAGGER_PUBLISHER_API_TOKEN }}@github.com/hmcts/cnp-api-docs.git
```
timja, line 32: the critical change was using a username here, it didn't seem to want to work with token only.

src/main/resources/application.yaml
```diff
@@ -65,6 +65,9 @@ idam:
     username: ${WA_SYSTEM_USERNAME:some_user@hmcts.net}
     password: ${WA_SYSTEM_PASSWORD:password}
 
+springdoc:
+  writer-with-order-by-keys: true
```
timja, line 69: previously on every commit a new spec was being published

.github/workflows/swagger.yml
```diff
@@ -4,35 +4,29 @@ on:
   push:
     branches:
       - master
-  pull_request:
-    branches:
-      - master
```
timja, line 9: shouldn't publish API specs from master.

you could change it to validate on PR if you wanted to

PR #1026, opened 2023-04-02
Rwa 2101 new db

chrisdavecm:
https://tools.hmcts.net/jira/browse/RWA-2101

### Change description ###
Flexible server details

```
[ ] Yes
[x] No
```


infrastructure/variables.tf
```diff
@@ -45,3 +45,10 @@ variable "database_sku_capacity" {
   default = "8"
 }
 
+variable "business_area" {
+  default = "cft"
+}
```
raghera, line 50: Any ideas what this is for?  I can see it being used in the DB config so sounds like it is correct.

config/owasp/suppressions.xml
```diff
@@ -67,4 +67,16 @@
...
+      tmp till fixed
+    </notes>
+    <cve>CVE-2023-28708</cve>
+  </suppress>
+  <suppress>
+    <notes>
+      tmp till fixed
+    </notes>
+    <cve>CVE-2023-20861</cve>
+  </suppress>
```
raghera, line 81: Not a blocker but we ought to put an end date on the suppression.

infrastructure/main.tf
```diff
@@ -48,6 +53,62 @@ module "wa_task_management_api_database" {
...
+    {
+      name : var.postgresql_database_name
+    }
+  ]
+  pgsql_version      = 14
+  common_tags        = local.common_tags
+
+  admin_user_object_id = var.jenkins_AAD_objectId
+
+}
```
raghera, line 111: This all looks good.
May be a question for PlatOps but will all the above config stay the same for when we want to go to prod or were there some temporary references to get it working in perftest?

raghera (approved):

charts/wa-task-management-api/values.preview.template.yaml
```diff
@@ -16,13 +16,17 @@ java:
...
+          alias: POSTGRES_USERNAME
   environment:
     CAMUNDA_URL: "http://camunda-api-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal/engine-rest"
     IDAM_URL: "https://idam-api.{{ .Values.global.environment }}.platform.hmcts.net"
     S2S_URL: "http://rpe-service-auth-provider-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal"
     OPEN_ID_IDAM_URL: "https://idam-web-public.{{ .Values.global.environment }}.platform.hmcts.net"
     ROLE_ASSIGNMENT_URL: "http://am-role-assignment-service-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal"
-    POSTGRES_HOST: "{{ .Release.Name }}-postgresql"
+    POSTGRES_HOST: cft-task-postgres-db-flexible-{{ .Values.global.environment }}.postgres.database.azure.com
     POSTGRES_NAME: "{{ .Values.postgresql.auth.database}}"
```
dharmendrak, line 30: this is wrong, we need to setup flexible db or use existing db for preview

PR #1077, opened 2023-04-27
Remove calculated dates from calculated dates output (RWA-2520)

dharmendrak:

[RWA-2520](https://tools.hmcts.net/jira/browse/RWA-2520)

### Change description ###

Remove calculated dates from calculated dates output

```
[ ] Yes
[ ] No
```


rasikasri (approved):

BenNzewi (commented):
Does this mean we no longer require the calculatedDates to be configured?

PR #1093, opened 2023-05-15
Rwa 2304

Lee365:
https://tools.hmcts.net/jira/browse/RWA-2304

### Change description ###
RWA-2304_6_7_8_9

```
[ ] Yes
[X] No
```


src/main/resources/application.yaml
```diff
@@ -139,6 +139,8 @@ launchDarkly:
   connectionTimeout: 5000
   socketTimeout: 5000
 
+environment: ${ENVIRONMENT:local}
```
rasikasri, line 142: you need to set this env variable in chart file to read the values of the other environments. Otherwise it will always set to local.
Like this - ENVIRONMENT: "{{ .Values.global.environment }}"

rasikasri (changes requested):

src/main/resources/application.yaml
Lee365: Good point. Added to values.yaml

rasikasri (approved):

PR #1109, opened 2023-06-14
RWA-2085: Subscription resilience

rasikasri:

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/resources/dbreplica/migration/V1.0.5__task_assignments_plsql.sql
```diff
@@ -0,0 +1,120 @@
...
+as $function$
+begin
+  perform cft_task_db.add_task_assignments(new.task_id);
+return new;
+end $function$;
+
+DROP TRIGGER IF EXISTS trg_on_task_history_insert ON cft_task_db.task_history;
+--
+CREATE TRIGGER trg_on_task_history_insert after insert on cft_task_db.task_history
+                                                           for each row execute function cft_task_db.on_task_history_insert();
```
raghera, line 119: I think this trigger would be better after insert on cft_task_db.tasks.  It is at this point that technically the replication has completed since we are inserting into the task_history _before_ it is inserted into the replica Tasks table.

src/main/resources/dbreplica/migration/V1.0.5__task_assignments_plsql.sql
```diff
@@ -0,0 +1,120 @@
...
+        or ((COALESCE(l_assignee,'') = '') and (l_update_action in ('Unassign', 'Unclaim', 'AutoUnassign'))) then
+
+        update cft_task_db.task_assignments
+            set  assignment_end =  l_updated,
+                 assignment_end_reason = l_end_reason
+            where task_id = l_task_id and assignment_end is null and assignment_start < l_updated;
+    end if;
+
+return l_update_id;
+end $function$;
```
raghera, line 102: Looks good.  So if it is a first assignment then there would only be an insert of a new assignment.  Otherwise it either updates the previous one and inserts the new one or updates the existing one depending on the action.

There's probably a bunch of tests to run through but the logic makes sense.

src/main/resources/dbreplica/migration/V1.0.4__reportable_task_plsql.sql
```diff
@@ -0,0 +1,202 @@
...
+      l_task.jurisdiction, l_task.region, l_task.location, l_task.business_context,
+      l_task.termination_reason, l_task.created, l_task.last_updated_user, l_task.last_updated_timestamp,
+      l_task.last_updated_action, l_task.created::DATE, l_task.due_date_time::date,
+      l_task.last_updated_timestamp::date)
+    returning update_id into l_update_id;
+end if;
+
+if l_result_action = 'update' then
+    select 1 into l_update_lock from cft_task_db.reportable_task
+    where reportable_task.task_id = l_task.task_id FOR UPDATE;
```
raghera, line 67: Probably need to talk about this a bit.  What happens in the case that the task_history has multiple rows but there is no record in the reportable_task table?  Looks like here we are only making an update/insert based on the current update?  Does that work in the above scenario?

raghera (commented):
Take a look.  Mostly really good just want to check the reportable_task logic with you.

PR #1122, opened 2023-07-18
Add feature toggle for delete tasks endpoint

MartinYSpasov:
https://tools.hmcts.net/jira/browse/LAU-653

### Change description ###
Add feature toggle for delete tasks endpoint

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/config/DeleteTaskFeatureToggleInterceptor.java
+++ b/src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/config/DeleteTaskFeatureToggleInterceptor.java
@@ -3,9 +3,11 @@ package uk.gov.hmcts.reform.wataskmanagementapi.config;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.web.servlet.HandlerInterceptor;
 
+import java.io.IOException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import static com.microsoft.applicationinsights.core.dependencies.http.HttpStatus.SC_SERVICE_UNAVAILABLE;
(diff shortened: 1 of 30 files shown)


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/config/DeleteTaskFeatureToggleInterceptor.java
```diff
@@ -0,0 +1,26 @@
...
+
+    @Autowired
+    private LaunchDarklyFeatureFlagProvider launchDarklyFeatureFlagProvider;
+
+    @Override
+    public boolean preHandle(final HttpServletRequest request,
+                             final HttpServletResponse response,
+                             final Object handler) {
+
+        return launchDarklyFeatureFlagProvider.getBooleanValue(
```
raghera, line 21: What would be the http response code if returning `false` (guessing there's a default it uses)?  I think setting it to something like a 501 or 503 would be good.

src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/config/DeleteTaskFeatureToggleInterceptorTest.java
```diff
@@ -0,0 +1,59 @@
...
+        when(launchDarklyFeatureFlagProvider.getBooleanValue(WA_DELETE_TASK_BY_CASE_ID,
+                "ccd-case-disposer",
+                "ccd-case-disposer@hmcts.net")).thenReturn(false);
+
+        final boolean isValidRequest = deleteTaskFeatureToggleInterceptor
+                .preHandle(httpServletRequest, httpServletResponse, object);
+
+        assertThat(isValidRequest).isEqualTo(false);
+    }
+}
```
raghera, line 59: Would be good to add an integration boot integration test you can check it works with the context loaded and see what the http response is.

raghera (commented):
Looks good in general. Couple of comments so take a look.

raghera (dismissed):

raghera (approved):

PR #1141, opened 2023-08-03
Correcting wal_level requirements

raghera:
wal_level on the replica DB does not need to be logical.  The default replica level is sufficient and this is already there.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/MIReportingService.java
+++ b/src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/MIReportingService.java
@@ -67,12 +67,9 @@ public class MIReportingService {
         Objects.requireNonNull(taskResourceRepository, "Primary Task DB repo is null.");
         Objects.requireNonNull(taskResourceRepository, "Replica Task DB repo is null.");
 
-        if (!taskResourceRepository.showWalLevel().equals(WAL_LEVEL)
-            || !reportableTaskRepository.showWalLevel().equals(WAL_LEVEL)) {
-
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/MIReportingService.java
```diff
@@ -67,12 +67,9 @@ public void logicalReplicationCheck() {
         Objects.requireNonNull(taskResourceRepository, "Primary Task DB repo is null.");
         Objects.requireNonNull(taskResourceRepository, "Replica Task DB repo is null.");
 
-        if (!taskResourceRepository.showWalLevel().equals(WAL_LEVEL)
-            || !reportableTaskRepository.showWalLevel().equals(WAL_LEVEL)) {
-
-            log.error("WAL LEVEL for primary DB; {}, replicaDB: {}.  These must be set to logical",
-                taskResourceRepository.showWalLevel(),
-                reportableTaskRepository.showWalLevel());
+        if (!taskResourceRepository.showWalLevel().equals(WAL_LEVEL)) {
```
kvvasudev, line 70: May be Rename of Constant WAL_LEVEL to something like WAL_LEVEL_LOGICAL may make it more readable.

kvvasudev (approved):

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/MIReportingService.java
raghera: 👍

PR #1142, opened 2023-08-07
RWA-2707 alert to notify of task deletion failure

rapolaskaseliscgi:

https://tools.hmcts.net/jira/browse/RWA-2707

### Change description ###

Added task deletion failure alert 

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/alerts.tf
+++ b/infrastructure/alerts.tf
@@ -1,0 +1,35 @@
+locals {
+  alert_resource_group_name = "wa-${var.env}"
+}
+
+module "wa-task-disposal-action-group" {
+  source                 = "git@github.com:hmcts/cnp-module-action-group"
+  location               = "global"
+  env                    = var.env
+  resourcegroup_name     = local.alert_resource_group_name
+  action_group_name      = "${var.application_name}-${var.env}-ag"
+  short_name             = "dispr-alert"
+  email_receiver_name    = "WA Task Deletion Failure Alert"
+  email_receiver_address = data.azurerm_key_vault_secret.wa-support-email.value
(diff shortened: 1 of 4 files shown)


infrastructure/alerts.tf
```diff
@@ -0,0 +1,35 @@
...
+  location               = "global"
+  env                    = var.env
+  resourcegroup_name     = local.alert_resource_group_name
+  action_group_name      = "${var.application_name}-${var.env}-ag"
+  short_name             = "dispr-alert"
+  email_receiver_name    = "WA Task Deletion Failure Alert"
+  email_receiver_address = data.azurerm_key_vault_secret.wa-support-email.value
+}
+
+module "task-deletion-failure-alert" {
```
raghera, line 16: In fact we have put all our alerts in our wa-shared-infrastructure repo to keep them in once place.  Can you move this to there please?
https://github.com/hmcts/wa-shared-infrastructure/blob/master/alerts.tf

There are many wa-support alerts which span across the product and not just task-management-api so wanted them in once place.

infrastructure/variables.tf
```diff
@@ -52,3 +52,11 @@ variable "business_area" {
 variable jenkins_AAD_objectId {}
 
 variable aks_subscription_id {}
+
+variable "enable_alerts" {
```
raghera, line 56: since we have a few different alerts (overall) better to name it more specifically like `enable-task-deletion-failure-alert` or `enable-task-deletion-alert`

raghera (changes requested):
Code looks fine just location.

PR #1147, opened 2023-08-17
Rwa 2664 1

kvvasudev:

https://tools.hmcts.net/jira/browse/RWA-2720

### Change description ###

Introduced subscription host and port as a config param

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/src/integrationTest/resources/application-integration.yaml
+++ b/src/integrationTest/resources/application-integration.yaml
@@ -34,7 +34,7 @@ idam:
     username: ${TEST_WA_LAW_FIRM_USERNAME:some_user@hmcts.net}
     password: ${TEST_WA_LAW_FIRM_PASSWORD:password}
 
-environment: ${ENVIRONMENT:arm}
+environment: ${ENVIRONMENT:local-arm-arch}
 
 role-assignment-service:
   url: http://localhost:8888
--- a/src/integrationTest/resources/application-integration.yaml
+++ b/src/integrationTest/resources/application-integration.yaml
@@ -34,6 +34,8 @@ idam:
     username: ${TEST_WA_LAW_FIRM_USERNAME:some_user@hmcts.net}
     password: ${TEST_WA_LAW_FIRM_PASSWORD:password}
 
+environment: ${ENVIRONMENT:arm}
+
 role-assignment-service:
   url: http://localhost:8888
 
--- a/src/main/resources/application-replica.yaml
+++ b/src/main/resources/application-replica.yaml
@@ -1,0 +1,2 @@
+primary:
+  publicationUrl: postgresql://${POSTGRES_CLUSTER_HOST:192.168.49.2}:${POSTGRES_PORT:5432}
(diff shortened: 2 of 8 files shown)


src/main/resources/application.yaml
```diff
@@ -77,6 +77,7 @@ replication:
 primary:
   username: ${POSTGRES_USERNAME:wa_user}
   password: ${POSTGRES_PASSWORD:wa_password}
+  cluster-host-name: postgresql://${POSTGRES_CLUSTER_HOST:192.168.49.2}:${POSTGRES_PORT:5432}
```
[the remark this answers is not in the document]
kvvasudev: Yes makes sense. Rename to publicationUrl and moved it to application-replicate.yml.

src/main/resources/application-replica.yaml
```diff
@@ -0,0 +1,2 @@
+primary:
+  publicationUrl: postgresql://${POSTGRES_CLUSTER_HOST:192.168.49.2}:${POSTGRES_PORT:5432}
```
raghera, line 2: Is there a way we can not have this available at all in prod?  I.e. a local dev profile only?  The whole idea about putting this in it's own file is to avoid it's usage everywhere except locally.  As I understand it the `replica` profile is what is run everywhere so this will effectively be wired in everywhere when it is not required.

raghera: Maybe you needed it everywhere?

raghera: replicaSubscriptionUrl

src/main/resources/application.yaml
```diff
@@ -74,6 +74,7 @@ springdoc:
 replication:
   username: ${REPLICATION_USERNAME:repl_user}
   password: ${REPLICATION_PASSWORD:repl_password}
+  subscriptionUrl: postgresql://${POSTGRES_CLUSTER_HOST:192.168.49.2}:${POSTGRES_PORT:5432}
```
raghera, line 77: It's not essential but would be good to have the host as `localhost`

src/main/resources/application.yaml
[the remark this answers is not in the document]
kvvasudev: moved it back to application.yml

src/main/resources/application-replica.yaml
kvvasudev: Renamed.

src/main/resources/application.yaml
kvvasudev: changed the default now back to localhost.

src/integrationTest/resources/application-integration.yaml
```diff
@@ -34,6 +34,8 @@ idam:
     username: ${TEST_WA_LAW_FIRM_USERNAME:some_user@hmcts.net}
     password: ${TEST_WA_LAW_FIRM_PASSWORD:password}
 
+environment: ${ENVIRONMENT:arm}
```
raghera, line 37: Should this be `arm` by default?  Surely this would fail for anyone where they don't need minikube tunnel?  Think the default could be just `local` and then an m1/2 Arm arch could be named `local-arm-arch` so it is more descriptive.

raghera: Then we need to update the Readme that this is what you'd have to supply as an env var if you're using an Arm arch.

raghera (approved):

PR #1184, opened 2023-10-26
RWA-2905 - Added SQL data fixes

SammiChong:
https://tools.hmcts.net/jira/browse/RWA-2905

### Change description ###

```
[ ] Yes
[x ] No
```


src/main/resources/scripts/rwa_2905_check_and_update.sql
```diff
@@ -0,0 +1,42 @@
...
+select count(*), task_name  from cft_task_db.tasks where role_category is null and jurisdiction = 'CIVIL' and task_name in ('Legal Advisor Small Claims Track Directions') group by task_name ;
+--  3	  Legal Advisor Small Claims Track Directions
+
+
+-- IMPLEMENTATION STEPS
+-----------------------------
+
+update cft_task_db.tasks set role_category = 'ADMIN' where role_category is null and jurisdiction = 'CIVIL' and task_name in ('Confirm Case Offline', 'Transfer Case Offline');
+update cft_task_db.tasks set role_category = 'JUDICIAL' where role_category is null and jurisdiction = 'CIVIL'  and task_name in ('Directions after Judgment (Damages)', 'Fast Track Directions', 'Small Claims Track Directions');
+update cft_task_db.tasks set role_category = 'LEGAL_OPERATIONS' where role_category is null and jurisdiction = 'CIVIL' and task_name in ('Legal Advisor Small Claims Track Directions');
```
raghera, line 35: Just to note that we should check the number of rows updated is the same as the number expected to be updated from the prechecks.

src/main/resources/scripts/rwa_2905_check_and_update.sql
```diff
@@ -0,0 +1,42 @@
+-- PRE DEPLOYMENT DATA PREPARATION STEPS
```
raghera, line 1: Rename this file?  Maybe `rwa-2905_update_civil_role_categories`?

src/main/resources/scripts/rwa_2905_rollback.sql
```diff
@@ -0,0 +1,188 @@
+update cft_task_db.tasks set role_category = null where task_id in (
```
raghera, line 1: Name this file the same as the one making the change and add a `_rollback` suffix.

src/main/resources/scripts/rwa_2905_check_and_update.sql
```diff
@@ -0,0 +1,42 @@
...
+-----------------------------------------
+select task_id from cft_task_db.tasks where role_category is null and jurisdiction = 'CIVIL';
+--  take a backup of all the task_id's
```
raghera, line 4: 👍 
Might have to get this at the time of execution.
Although a very small possibility and possibly negligible risk,  it's not guaranteed that even if the number of nulls are the same as today, that the task_ids are the same as in the list in your rollback.

raghera (approved):

PR #1187, opened 2023-11-02
RWA-2857: MI - collate mark and refresh

kvvasudev:
### Change description ###

Collate Mark and Refresh

```
[ ] Yes
[ x ] No
```


raghera (approved):

raghera (approved):
Changes look good.  I just wonder if the time window of 100 millis for verifying the time is going to be consistent enough when running in jenkins.  Hopefully it is so let's see.

PR #1193, opened 2023-11-27
RWA-2715: Shouldn't complete already terminated tasks

kvvasudev:

https://tools.hmcts.net/jira/browse/RWA-2715

### Change description ###

Completed tasks not transitioning to a terminated state

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskManagementService.java
+++ b/src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskManagementService.java
@@ -552,18 +552,12 @@ public class TaskManagementService {
             //check the state, if not complete, complete
             completeCamundaTask(taskId, taskHasCompleted);
             //Commit transaction
-            if (isActive(state)) {
+            if (task.isActive(state)) {
                 task.setState(CFTTaskState.COMPLETED);
                 setTaskActionAttributes(task, userId, TaskAction.COMPLETED);
                 cftTaskDatabaseService.saveTask(task);
             }
         }
-    }
-
-    private static boolean isActive(CFTTaskState state) {
-        return !(state.equals(CFTTaskState.TERMINATED)
-            || state.equals(CFTTaskState.COMPLETED)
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskManagementService.java
```diff
@@ -541,7 +541,8 @@ public void completeTask(String taskId, AccessControlResponse accessControlRespo
 
         //Lock & update Task
         TaskResource task = findByIdAndObtainLock(taskId);
-        taskHasCompleted = task.getState() == CFTTaskState.COMPLETED;
+        taskHasCompleted = task.getState() != null
+            && (task.getState().equals(CFTTaskState.COMPLETED) || task.getState().equals(CFTTaskState.TERMINATED));
```
raghera, line 545: I think there's a slight edge case here which we should cater for.  If a Task is TERMINATED then it could have been completed or cancelled previously and this info is in the termination_reason field.  So how do we deal with the case where a request is to complete a Task but it has been Cancelled?  I think the current behaviour doesn't cater for this also.
I think it should be an error although can discuss the details.

raghera (commented):
See comment

raghera (approved):

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/TaskManagementService.java
```diff
@@ -541,17 +541,22 @@ public void completeTask(String taskId, AccessControlResponse accessControlRespo
 
         //Lock & update Task
         TaskResource task = findByIdAndObtainLock(taskId);
-        taskHasCompleted = task.getState() == CFTTaskState.COMPLETED;
+        CFTTaskState state = task.getState();
+        taskHasCompleted = state != null
```
BenNzewi, line 545: New line of code and is partially covered.  May need to include terminationReason in test where cftTask.state = COMPLETED.

PR #1238, opened 2024-03-15
RWA-3142 added a condition to not add additionalProperties if not con…

skandala-cgi:
https://tools.hmcts.net/jira/browse/RWA-3119

### Change description ###
updated to set to null for empty additional properties

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CFTTaskMapperTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CFTTaskMapperTest.java
@@ -2449,6 +2449,14 @@ class CFTTaskMapperTest {
 
     }
 
+    @Test
+    void do_not_reconfigure_additional_attributes_fields_if_empty_list_returned() {
+        TaskResource taskResource = createTaskResource();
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/hmcts/reform/wataskmanagementapi/services/CFTTaskMapperTest.java
```diff
@@ -2449,6 +2449,14 @@ void reconfigure_config_additional_attributes_dmn_fields() {
 
     }
 
+    @Test
+    void do_not_reconfigure_additional_attributes_fields_if_empty_list_returned() {
+        TaskResource taskResource = createTaskResource();
+        List<ConfigurationDmnEvaluationResponse> additionalPropertyDMNResponse = new ArrayList<>();
+        cftTaskMapper.reconfigureAdditionalTaskAttribute(taskResource, additionalPropertyDMNResponse);
+        assertNull(taskResource.getAdditionalProperties());
+    }
```
raghera, line 2458: This is ok for now to address the problem in prod.  However, we should review the functional tests to ensure we are not changing the json response when there are empty additional properties.  It's not a breaking change in the API as it's an agreed field but could break some clients if coded badly.

raghera (approved):

PR #1242, opened 2024-03-20
RWA-3173: MI Reporting fix blanks

kvvasudev:

https://tools.hmcts.net/jira/browse/RWA-3173

### Change description ###

Replica Task Assignments gate for incomplete history

```
[ ] Yes
[ x ] No
```


src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/services/MIReplicaReportingServiceTest.java
```diff
@@ -319,6 +319,75 @@ void should_save_task_and_get_transformed_state_label_from_reportable_task(
                 });
     }
 
+    @ParameterizedTest
+    @CsvSource(value = {
+        "ASSIGNED,AutoAssign,true,true",
+        "UNASSIGNED,Configure,false,true",
+        "COMPLETED,AutoAssign,false,false",
+        "CANCELLED,Configure,false,false"
```
raghera, line 327: Are there a few more permutations here?  e.g. ASSIGNED,AutoAssign,false,false?
Could we also have cases where the state is COMPLETED and there is already a report there?

src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/services/MIReplicaReportingServiceTest.java
```diff
@@ -891,6 +960,27 @@ private TaskResource createAndSaveThisTask(String taskId, String taskName,
...
+        taskResource.setCreated(OffsetDateTime.parse("2023-03-23T20:15:45.345875+01:00"));
+        taskResource.setPriorityDate(OffsetDateTime.parse("2023-03-26T20:15:45.345875+01:00"));
+        taskResource.setLastUpdatedAction(lastAction);
+        taskResource.setLastUpdatedTimestamp(OffsetDateTime.parse("2023-03-29T20:15:45.345875+01:00"));
+        taskResource.setJurisdiction("someJurisdiction");
+        taskResource.setLocation("someLocation");
+        taskResource.setRoleCategory("someRoleCategory");
+        taskResource.setAssignee(assignee);
+        return taskResourceRepository.save(taskResource);
+    }
```
raghera, line 982: As a general rule it's better to put test object creators into their own class so it can be re-used by many tests.  
Private methods like this just mean that someone else will probably write something very similar if not the same in another class.

src/main/resources/dbreplica/migration/V1.0.32__task_assignments_plsql.sql
```diff
@@ -0,0 +1,66 @@
...
+l_assignee      text;
+l_update_action text;
+l_updated       TIMESTAMP;
+l_state         text;
+l_jurisdiction  text;
+l_location      text;
+l_role_category text;
+l_task_name     text;
+l_new_task      boolean;
+l_valid_history boolean DEFAULT false;
```
raghera, line 20: Is this provided from elsewhere or is this always false?

kvvasudev: Hello Ravi, Default is false but I am setting it to true in line 45.

src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/services/MIReplicaReportingServiceTest.java
kvvasudev: Hello Ravi, In this test, I am not maintaining state transitions, but directly creating tasks with final State and final Update Action. Just to test if the task_assignments gates are working.

raghera (dismissed):

SammiChong (approved):

PR #1256, opened 2024-05-02
RWA-3307 Race condition with async Reconfigurations and Terminations

skandala-cgi:

https://tools.hmcts.net/jira/browse/RWA-3307

### Change description ###

Race condition with async Reconfigurations and Terminations

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/repository/TaskResourceRepository.java
+++ b/src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/repository/TaskResourceRepository.java
@@ -80,7 +80,7 @@ public interface TaskResourceRepository extends CrudRepository<TaskResource, Str
         List<CFTTaskState> states, OffsetDateTime reconfigureRequestTime);
 
     @Lock(LockModeType.PESSIMISTIC_WRITE)
-    @QueryHints({@QueryHint(name = LOCK_TIMEOUT_STR, value = "0")})
+    @QueryHints({@QueryHint(name = LOCK_TIMEOUT_STR, value = "5000")})
     @Transactional
     @NonNull
     @Query("select t from tasks t where t.taskId = :id and state in (:states)")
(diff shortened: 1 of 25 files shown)


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/operation/ExecuteTaskReconfigurationService.java
```diff
@@ -124,8 +124,8 @@ private List<String> reconfigureTasks(List<String> taskIds, List<TaskResource> s
                 try {
                     log.info("Re-configure task-id {}", taskId);
                     Optional<TaskResource> optionalTaskResource = cftTaskDatabaseService
-                        .findByIdAndObtainPessimisticWriteLock(taskId);
-
+                         .findByIdAndStateInObtainPessimisticWriteLock(taskId, List.of(CFTTaskState.ASSIGNED,
+                                                                                      CFTTaskState.UNASSIGNED));
                     if (optionalTaskResource.isPresent()) {
```
raghera, line 129: In the case where we initially want to Reconfigure a taskId but then don't want to execute because the state has changed we should output logs giving at least a description which includes the taskId.  It may even be beneficial to include the caseId and state of the Task.
E.g. For <OperationName> did not execute for Task Resource: taskId: {}, caseId: {}, state {}

This will allow us to track in case we need to.

However this may mean that you have to either go and get more info from the Task record with another DB query or your original DB query returns the full Task.

We've seen in production that this is a rare use case so querying the DB again would be acceptable.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/repository/TaskResourceRepository.java
```diff
@@ -77,6 +79,14 @@ List<TaskResource> findByCaseIdInAndStateInAndReconfigureRequestTimeIsNull(
     List<TaskResource> findByStateInAndReconfigureRequestTimeGreaterThan(
         List<CFTTaskState> states, OffsetDateTime reconfigureRequestTime);
 
+    @Lock(LockModeType.PESSIMISTIC_WRITE)
+    @QueryHints({@QueryHint(name = LOCK_TIMEOUT_STR, value = "0")})
```
raghera, line 83: So I think we should think about extending the lock timeout here.  The behaviour we want is to wait for any other threads to complete an update before we run our query.  I think a value of 0 would throw an exception straight away which may not be what we want e.g. if the previous thread is a Task Assignment rather than a Cancellation.  I think 5000 millis is enough to start with.

PR #1257, opened 2024-05-07
RWA-3201 Update ET Service name

skandala-cgi:
https://tools.hmcts.net/jira/browse/RWA-3201

### Change description ###
Update ET Service name

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/dbreplica/migration/V1.0.33__reportable_task_plsql.sql
+++ b/src/main/resources/dbreplica/migration/V1.0.33__reportable_task_plsql.sql
@@ -1,0 +1,324 @@
+CREATE OR REPLACE FUNCTION cft_task_db.add_reportable_task(l_task_id text)
+ RETURNS bigint
+ LANGUAGE plpgsql
+AS $function$
+
+declare
+l_update_id                       bigint;
+l_new_task                        boolean;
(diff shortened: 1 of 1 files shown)


src/main/resources/dbreplica/migration/V1.0.33__reportable_task_plsql.sql
```diff
@@ -0,0 +1,324 @@
...
+            else l_role_category
+        end;
+
+    l_jurisdiction_label =
+        case l_jurisdiction IS NOT NULL
+            when (l_jurisdiction='PRIVATELAW')  then 'Private Law'
+            when (l_jurisdiction='CIVIL')  then 'Civil'
+            when (l_jurisdiction='IA')  then 'Immigration and Asylum'
+            when (l_jurisdiction='PUBLICLAW')  then 'Public Law'
+            when (l_jurisdiction='EMPLOYMENT')  then 'Employment'
```
raghera, line 121: As discussed please check that SSCS are happy for their label to be `SSCS` rather than something else as we can add their mapping here too.

raghera (approved):
PR is fine.  Just a question about SSCS which would be good to add in this PR if possible.

PR #1259, opened 2024-05-13
RWA-2482: Non Prod task cleanup scripts

adityadwadasi:

https://tools.hmcts.net/jira/browse/RWA-2482 

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
```diff
@@ -0,0 +1,82 @@
...
+
+CREATE OR REPLACE FUNCTION task_cleanup_primary()
+RETURNS VOID
+LANGUAGE plpgsql
+AS $function$
+DECLARE
+    id text;
+BEGIN
+    -- Loop through the rows of the tasks table
+    FOR id IN select task_id from cft_task_db.tasks WHERE EXTRACT(YEAR FROM created) in (2021) LOOP
```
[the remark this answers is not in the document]
raghera: If we are doing by time it makes sense also to order them by oldest first and maybe set a max limit so we cannot accidentally have 20 million tasks being deleted (unlikely as it is right now).  You could say delete the first 10 000 only.  If more exist then we can re-run the same again.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
```diff
@@ -0,0 +1,82 @@
...
+        END IF;
+
+        -- delete from tasks
+        DELETE FROM cft_task_db.tasks WHERE task_id = id;
+
+        IF EXISTS (SELECT 1 FROM cft_task_db.tasks WHERE task_id = id) THEN
+            RAISE NOTICE 'Issue with the taskId: % in tasks table', id;
+            RETURN;
+        END IF;
+    END LOOP;
```
raghera, line 36: Might have to test this but usually functions run in one transaction and so it all completes or it all fails.  Since you are looping through many records what if 10 pass and the last one fails?  Will the whole thing rollback?

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
```diff
@@ -0,0 +1,82 @@
...
+        IF EXISTS (SELECT 1 FROM cft_task_db.sensitive_task_event_logs where task_id = id) THEN
+            RAISE NOTICE 'Issue with the taskId: % in sensitive_task_event_logs table', id;
+        END IF;
+
+        -- delete from task_roles
+        delete from cft_task_db.task_roles where task_id = id;
+
+        IF EXISTS (SELECT 1 FROM cft_task_db.task_roles where task_id = id) THEN
+            RAISE NOTICE 'Issue with the taskId: % in task_roles table', id;
+        END IF;
```
raghera, line 27: If you can't delete the role then the task delete will fail so probably want to abort that task_id otherwise you'll get an exception which you will have to handle.  Otherwise the whole thing would stop.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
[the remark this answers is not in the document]
adityadwadasi: I added the limit, modified the queries to accept date/time and updated the function name as well Ravi.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
```diff
@@ -0,0 +1,82 @@
...
+
+CREATE OR REPLACE FUNCTION task_cleanup_replica()
+RETURNS VOID
+LANGUAGE plpgsql
+AS $function$
+DECLARE
+    id text;
+BEGIN
+    -- Loop through the rows of the tasks table
+    FOR id IN select task_id from cft_task_db.tasks WHERE EXTRACT(YEAR FROM created) in (2021) LOOP
```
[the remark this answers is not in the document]
adityadwadasi: I added the limit, modified the queries to accept date/time and updated the function name as well Ravi.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
```diff
@@ -0,0 +1,82 @@
...
+
+        -- Check if any more rows exist in tasks
+        IF EXISTS (SELECT 1 FROM cft_task_db.tasks where task_id = id) THEN
+            RAISE NOTICE 'Issue with the taskId: % in tasks table', id;
+        END IF;
```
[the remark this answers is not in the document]
adityadwadasi: I made the changes accordingly. Since the tasks are being deleted through triggers, Instead looking for tasks  in the replica task table, i am taking distinct tasks from the task_history table.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
adityadwadasi: I moved the return to the end of the function. This scenario is avoided.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
adityadwadasi: Made the necessary changes. If any issue occurs while deleting in the child tables, i am skipping that task ID deletion in the tasks table.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
```diff
@@ -0,0 +1,92 @@
...
+
+CREATE OR REPLACE FUNCTION cft_task_db.task_cleanup_between_dates_primary()
+RETURNS VOID
+LANGUAGE plpgsql
+AS $function$
+DECLARE
+    task_id_var text;
+BEGIN
+    -- Loop through the rows of the tasks table
+    FOR task_id_var IN SELECT task_id FROM cft_task_db.tasks WHERE created >= '<From Date and time>' AND created <= '<To Date and time>' limit 10000 LOOP
```
raghera, line 13: maybe better to pass the from and to date and limits as params into the function?

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
```diff
@@ -0,0 +1,92 @@
...
+
+        -- delete from sensitive_task_event_logs
+        DELETE FROM cft_task_db.sensitive_task_event_logs WHERE task_id = task_id_var;
+
+        IF EXISTS (SELECT 1 FROM cft_task_db.sensitive_task_event_logs WHERE task_id = task_id_var) THEN
+            RAISE NOTICE 'Issue with the taskId: % in sensitive_task_event_logs table. Skipping next steps', task_id_var;
+            CONTINUE;
```
raghera, line 20: If there is no sensitive log entry it should not stop the reset of the deletion.  Can be zero or many sensitive logs per task_id.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
raghera: 👍

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
```diff
@@ -0,0 +1,92 @@
...
+        IF EXISTS (SELECT 1 FROM cft_task_db.task_assignments WHERE task_id = task_id_var) THEN
+          RAISE NOTICE 'Issue with the taskId: % in task_assignments table', task_id_var;
+        END IF;
+
+        -- delete from task_history
+        DELETE FROM cft_task_db.task_history WHERE task_id = id;
+
+        IF EXISTS (SELECT 1 FROM cft_task_db.task_history WHERE task_id = task_id_var) THEN
+            RAISE NOTICE 'Issue with the taskId: % in task_history table', task_id_var;
+        END IF;
```
raghera, line 85: If something above fails then we don't want to delete from history.  The reason is we are using the history as the master of whether something still exists or not in the Replica so this is the last thing we should delete and only if everything else has definitely been deleted.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
raghera: Not sure if we are on the same page.  Can discuss.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
raghera: order by time and clear the oldest first

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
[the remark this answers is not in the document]
adityadwadasi: I added perameter to accept created from and to dates.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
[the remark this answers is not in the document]
adityadwadasi: I added perameter to accept created from and to dates.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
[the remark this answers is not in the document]
adityadwadasi: I made changes to skip the task deletion from child tables if the tasks is till there in the replica task table.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
adityadwadasi: As discussed I added commit at the end the loop. It will commit all the successful tasks until any exception.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
adityadwadasi: Added orderby and added created to and from params

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
```diff
@@ -0,0 +1,92 @@
...
+CREATE OR REPLACE FUNCTION cft_task_db.task_cleanup_between_dates_replica()
+RETURNS VOID
+LANGUAGE plpgsql
+AS $function$
+DECLARE
+    task_id_var text;
+BEGIN
+
+    -- Loop through the rows of the tasks table
+    FOR task_id_var IN SELECT DISTINCT task_id FROM cft_task_db.task_history WHERE created >= '<From Date and time>' AND created <= '<To Date and time>' limit 10000 LOOP
```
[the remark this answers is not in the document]
adityadwadasi: Added orderby and added created to and from params

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
adityadwadasi: I am skipping the task ID from deleting from history table if any data is still there in the child tables for that purticular task.

raghera (dismissed):
Looks good Aditya!
I think in future we should add some automated tests but it's ok for now since we are just trying to delete records in non-prod which this does.  I note you have also tested this locally to verify.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
```diff
@@ -0,0 +1,101 @@
...
+
+CREATE OR REPLACE FUNCTION cft_task_db.task_cleanup_between_dates_primary(create_from TIMESTAMP,create_to TIMESTAMP)
+RETURNS VOID
+LANGUAGE plpgsql
+AS $function$
+DECLARE
+    task_id_var TEXT;
+BEGIN
+    -- Loop through the rows of the tasks table
+    FOR task_id_var IN SELECT task_id FROM cft_task_db.tasks WHERE created >= create_from AND created <= create_to order by created limit 2000 LOOP
```
raghera, line 16: Is this limit what you think is correct following the deletions you did?

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
```diff
@@ -0,0 +1,101 @@
+--
+-- RWA-2482 cleanup old tasks from primary task database
+--
+--Example for calling the function
+--select cft_task_db.task_cleanup_between_dates_primary('2023-04-04 11:00'::TIMESTAMP,'2023-04-04 11:01'::TIMESTAMP);
+--
+
+CREATE OR REPLACE FUNCTION cft_task_db.task_cleanup_between_dates_primary(create_from TIMESTAMP,create_to TIMESTAMP)
```
raghera, line 8: sorry, should have seen this before but should it not be `created_from` and `created_to`?

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
```diff
@@ -0,0 +1,101 @@
...
+$function$
+
+--
+-- RWA-2482 cleanup old tasks from replica task database
+--
+--Example for calling the function
+--select cft_task_db.task_cleanup_between_dates_replica('2023-04-04 11:00'::TIMESTAMP,'2023-04-04 11:01'::TIMESTAMP);
+--
+
+CREATE OR REPLACE FUNCTION cft_task_db.task_cleanup_between_dates_replica(create_from TIMESTAMP,create_to TIMESTAMP)
```
raghera, line 55: same comment about the created naming.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
adityadwadasi: Yes Ravi... If the limit is more than 2000 the triggers are chocking the replica DB. CPU usage is going beyond 95%.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
adityadwadasi: Apologies. I missed it too. I will update the names.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
adityadwadasi: Apologies. I missed it too. I will update the names.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
raghera: Ok sounds good then.

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
```diff
@@ -0,0 +1,101 @@
...
+
+CREATE OR REPLACE FUNCTION cft_task_db.task_cleanup_between_dates_primary(created_time_from TIMESTAMP,created_time_to TIMESTAMP)
+RETURNS VOID
+LANGUAGE plpgsql
+AS $function$
+DECLARE
+    task_id_var TEXT;
+BEGIN
+    -- Loop through the rows of the tasks table
+    FOR task_id_var IN SELECT task_id FROM cft_task_db.tasks WHERE created >= create_from AND created <= create_to order by created limit 2000 LOOP
```
raghera, line 16: Don't think this will work.  You have to update the names of the params you're using.

raghera (approved):

src/main/resources/scripts/cleanup/non_prod_primary_and_replica_data_cleanup.sql
adityadwadasi: Oh no... I missed this... I made the changes now Ravi.... Please verify...

PR #1268, opened 2024-07-04
Revert "RWA-3485: Define DockerImageName"

SammiChong:
Reverts hmcts/wa-task-management-api#1266

Build failing


raghera (approved):
Logs look like they're pulling from the ACR and this looks like a less invasive way to do this so nice work!
Get Ben to double check the jenkins log as part of this QA in case we missed anything.

BenNzewi (approved):

PR #1279, opened 2024-07-29
RWA-3617: Set pgsql_storage_mb

SammiChong:
https://tools.hmcts.net/jira/browse/RWA-3617

### Change description ###

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -71,6 +71,7 @@ module "wa_task_management_api_database_flexible" {
   ]
 
   pgsql_version      = 14
+  pgsql_storage_mb   = var.pgsql_storage_mb
   common_tags        = local.common_tags
 
   admin_user_object_id = var.jenkins_AAD_objectId
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -70,9 +70,8 @@ module "wa_task_management_api_database_flexible" {
     }
   ]
 
-  pgsql_version    = 14
-  pgsql_storage_mb = var.pgsql_storage_mb
-  common_tags      = local.common_tags
+  pgsql_version = 14
+  common_tags   = local.common_tags
 
   admin_user_object_id = var.jenkins_AAD_objectId
 
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -51,6 +51,7 @@ module "wa_task_management_api_database_flexible" {
   product       = var.product
   component     = var.component
   name          = "${var.postgres_db_component_name}-postgres-db-flexible"
+  pgsql_sku     = var.pgsql_sku
   location      = var.location
   business_area = var.business_area
   env           = var.env
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -47,13 +47,13 @@ module "wa_task_management_api_database_flexible" {
     azurerm.postgres_network = azurerm.postgres_network
   }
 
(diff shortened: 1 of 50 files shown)


infrastructure/main.tf
```diff
@@ -47,13 +47,13 @@ module "wa_task_management_api_database_flexible" {
...
-  business_area      = var.business_area
-  env                = var.env
+  source        = "git@github.com:hmcts/terraform-module-postgresql-flexible?ref=master"
+  product       = var.product
+  component     = var.component
+  name          = "${var.postgres_db_component_name}-postgres-db-flexible"
+  location      = var.location
+  business_area = var.business_area
+  env           = var.env
   pgsql_databases = [
```
raghera, line 57: seems to be just formatting changes here?  Can you undo those please?

infrastructure/main.tf
```diff
@@ -83,21 +84,21 @@ module "wa_task_management_api_database_flexible_replica" {
...
   }
 
-  source             = "git@github.com:hmcts/terraform-module-postgresql-flexible?ref=master"
-  product            = var.product
-  component          = var.component
-  name               = "${var.postgres_db_component_name}-postgres-db-flexible-replica"
-  location           = var.location
-  business_area      = var.business_area
-  env                = var.env
+  source        = "git@github.com:hmcts/terraform-module-postgresql-flexible?ref=master"
```
raghera, line 87: same here for formatting.

infrastructure/variables.tf
```diff
@@ -49,6 +49,10 @@ variable "business_area" {
   default = "cft"
 }
 
-variable jenkins_AAD_objectId {}
+variable "pgsql_storage_mb" {
+  default = 65536
+}
```
raghera, line 54: This could change it everywhere?  Were we advised to do that?

infrastructure/main.tf
SammiChong: Looks like the bot automatically formatted it this way. https://github.com/hmcts/wa-task-management-api/pull/1279/commits/856534032cc781fb54d3a7cb5f1784fe08efce2c

infrastructure/variables.tf
SammiChong: This is the default size https://github.com/hmcts/terraform-module-postgresql-flexible?tab=readme-ov-file#input_pgsql_storage_mb so I'm defining it here and then setting the size it got upgraded to in Azure in demo.tfvars

raghera: The default should be set already in the terraform module so we shouldn't need to set it ourselves (you can prob find it in the module repo).  You should therefore just be able to override in demo where we want to change it?

Hopefully that will avoid all the destroy labels in prod etc.

infrastructure/main.tf
raghera: That's annoying it's made it worse

raghera: Can leave as is in that case.

infrastructure/variables.tf
SammiChong: When I tried to remove the default form the file I got a this error, "Error: Reference to undeclared input variable", looks like I'll have to have it set.

raghera (dismissed):
Approved but please co-ordinate deployment since there's some performance changes we want to get green first.

raghera (approved):

PR #1288, opened 2024-08-15
RWA 3193 mandatory fields add new flag part1

skandala-cgi:
https://tools.hmcts.net/jira/browse/RWA-3715

### Change description ###

Add new Launchdarkly Flag

```
[x] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/utils/TaskMandatoryFieldsValidator.java
```diff
@@ -0,0 +1,139 @@
...
+     * Checks if the task's jurisdiction is excluded.
+     *
+     * @param task the task to be checked
+     * @param excludedJurisdictionsArray the array of excluded jurisdictions
+     * @return true if the jurisdiction is excluded, false otherwise
+     */
+    private boolean isJurisdictionExcluded(TaskResource task, JsonNode excludedJurisdictionsArray) {
+        for (JsonNode jurisdiction : excludedJurisdictionsArray) {
+            if (jurisdiction.asText().equals(task.getJurisdiction())) {
+                return true;
```
raghera, line 95: Possibly a log of an excluded jurisdiction would be helpful for debug.  Over time there should be no exclusions would be good to know if for instance we end up with some new empty Role Categories how that happened.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/utils/TaskMandatoryFieldsValidator.java
```diff
@@ -0,0 +1,139 @@
...
+     * @return the JsonNode for the mandatory field check flag variant
+     * @throws IllegalArgumentException if the JSON string cannot be parsed
+     */
+    public JsonNode parseJson(String jsonString) {
+        try {
+            return new ObjectMapper().readTree(jsonString).get(MANDATORY_FIELD_CHECK_FLAG_VARIANT.stringValue());
+        } catch (JsonProcessingException e) {
+            throw new IllegalArgumentException("Mandatory flag jurisdictions mapping issue.", e);
+        }
+    }
```
raghera, line 114: Maybe this could be in another class (JsonParser if not one already?) and made more generic  e.g. public JsonNode getJsonNode(String jsonString, String nodeName) ?  

Also a minor consideration should we be using the ObjectMapper from the SpringContext or is there a problem with that?

src/main/resources/application.yaml
```diff
@@ -16,6 +16,8 @@ config:
   exclusiveAccessClients: ${TASK_MANAGEMENT_EXCLUSIVE_CLIENTS:wa_task_management_api,wa_task_monitor,wa_case_event_handler,wa_workflow_api}
   allowedJurisdictions: ${ALLOWED_JURISDICTIONS:ia,wa,sscs,civil,publiclaw,privatelaw,employment}
   allowedCaseTypes: ${ALLOWED_CASE_TYPES:asylum,wacasetype,sscs,civil,generalapplication,care_supervision_epo,prlapps,et_englandwales,et_englandwales_listings,et_englandwales_multiple,et_scotland,et_scotland_listings,et_scotland_multiple,et_admin,privatelaw_exceptionrecord,benefit}
+  taskMandatoryFieldCheckEnabled: ${MANDATORY_TASK_FIELD_CHECK_ENABLED:true}
```
raghera, line 19: Perhaps `false` is the safer default for now.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/utils/TaskMandatoryFieldsValidator.java
```diff
@@ -0,0 +1,139 @@
...
+                if (PropertyUtils.getProperty(task, field) == null
+                    || PropertyUtils.getProperty(task, field).toString().isBlank()) {
+                    errors.add(field + " cannot be null or empty");
+                }
+            } catch (Exception e) {
+                throw new IllegalArgumentException("Cannot find property value for field " + field, e);
+            }
+        }
+        if (!errors.isEmpty()) {
+            throw new ValidationException(String.join(", ", errors));
```
raghera, line 136: Something to consider here is that there are 2 flavours of mandatory field violations.  1) service team has not provided one though a config error 2) we have not provided one through our code.  The former is going to be what we are most interested in since with the latter some test is likely to fail.  Where the services have made a mistake we want to make sure we alert then asap and so we have to know it was a field which was their responsibility that was the problem (not our code).
Given this our model should reflect this distinction.  One can be used to alert the service but the other not.

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/utils/TaskMandatoryFieldsValidator.java
```diff
@@ -0,0 +1,150 @@
...
+            throw new ServiceMandatoryFieldValidationException(
+                String.join(", ", serviceSpecificErrors.toString()));
+        }
+        if (!tmSpecificErrors.isEmpty()) {
+            throw new ValidationException(String.join(", ", tmSpecificErrors.toString()));
+        }
+    }
+
+    private boolean isFieldNullOrEmpty(Object fieldValue) {
+        return fieldValue == null || fieldValue.toString().isBlank();
```
raghera, line 139: Will the Object.toString() ever be blank?  If it's not set won't it just be an object reference representation?  Maybe you have tested that scenario?

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/utils/TaskMandatoryFieldsValidator.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    /**
+     * Validates the mandatory fields of a task.
+     *
+     * @param task the task to be validated
+     * @throws ValidationException if any mandatory field is missing or invalid
+     * @throws IllegalArgumentException if a property value cannot be found
+     */
+    public void validateTaskMandatoryFields(TaskResource task) {
```
raghera, line 114: Is there a unit test for this?  I can't see anything called TaskMandatoryFieldsValidatorTest only an Integration Test.

skandala-cgi: added one now

src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/services/utils/TaskMandatoryFieldsValidator.java
skandala-cgi: If the value is empty string then fieldValue.toString().isBlank() is returning true.

raghera (approved):

src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/utils/TaskMandatoryFieldsValidatorIntegrationTest.java
```diff
@@ -0,0 +1,179 @@
...
+import java.util.List;
+import java.util.UUID;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+
+public class TaskMandatoryFieldsValidatorIntegrationTest extends SpringBootIntegrationBaseTest {
```
BenNzewi, line 32: General comment on the test method naming convention. Are we going with "given and no longer should"

src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/utils/TaskMandatoryFieldsValidatorIntegrationTest.java
```diff
@@ -0,0 +1,179 @@
...
+        task.setJurisdiction("WA");
+
+        TaskMandatoryFieldsValidator spyValidator = Mockito.spy(taskMandatoryFieldsValidator);
+        spyValidator.validate(task);
+
+        Mockito.verify(spyValidator, Mockito.never()).validateTaskMandatoryFields(task);
+    }
+
+    @Test
+    @DisplayName("should throw ValidationException when mandatory field is missing")
```
BenNzewi, line 133: Is this a duplicate of given_empty_mandatory_field_when_validate_then_throw_service_mandatory_field_validation_exception above?

src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/utils/TaskMandatoryFieldsValidatorIntegrationTest.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    @Test
+    @DisplayName("should validate successfully when all mandatory fields are present")
+    void given_all_mandatory_fields_present_when_validate_then_success() {
+        TaskResource task = getTaskResource(taskId);
+        assertDoesNotThrow(() -> taskMandatoryFieldsValidator.validate(task));
+    }
+
+    @Test
```
BenNzewi, line 60: Can a parameterised test allow the check for all our mandatory fields here?

BenNzewi (approved):

PR #1397, opened 2025-03-05
add query_work type

sonny-hmcts:

https://tools.hmcts.net/jira/browse/RWA-4156

### Change description ###

Added the work type 'query_work'

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskSearchControllerTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskSearchControllerTest.java
@@ -513,7 +513,8 @@ class PostTaskSearchControllerTest extends SpringBootIntegrationBaseTest {
                                + "decision_making_work, applications, priority, access_requests, "
                                + "error_management, review_case, evidence, follow_up, pre_hearing, post_hearing, "
                                + "intermediate_track_hearing_work, multi_track_hearing_work, "
-                               + "intermediate_track_decision_making_work, multi_track_decision_making_work]")
+                               + "intermediate_track_decision_making_work, multi_track_decision_making_work, "
+                               + "query_work]")
                 ));
     }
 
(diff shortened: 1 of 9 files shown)


skandala-cgi (approved):

src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskSearchControllerTest.java
```diff
@@ -513,7 +513,8 @@ void should_return_400_bad_request_when_work_types_did_not_exist() throws Except
                                + "decision_making_work, applications, priority, access_requests, "
                                + "error_management, review_case, evidence, follow_up, pre_hearing, post_hearing, "
                                + "intermediate_track_hearing_work, multi_track_hearing_work, "
-                               + "intermediate_track_decision_making_work, multi_track_decision_making_work]")
+                               + "intermediate_track_decision_making_work, multi_track_decision_making_work, "
+                               + "query_work]")
```
raghera, line 517: It would be good to add a test which uses the new Work Type as a filter attribute to the searches.  I expect there is already a test somewhere that does this with other Work Types so you may be able to extend it and use this work type.
This particular test is just checking that the 400 error includes query_work as one of the acceptable Work Types - so not using it as a filter.

src/main/resources/db/migration/V1.0.33__add_new_work_types.sql
```diff
@@ -0,0 +1,2 @@
+insert into work_types(work_type_id, label) values
+    ('query_work','Query work');
```
raghera, line 2: Did we get confirmation from the service that they are happy with the above label?  The above was our working assumption based on the others.

src/integrationTest/java/uk/gov/hmcts/reform/wataskmanagementapi/controllers/PostTaskSearchControllerTest.java
raghera: The same test above can verify that the new Work Type id is included in the search response.

raghera (commented):
Looks good, just a couple of comments.

src/main/resources/db/migration/V1.0.33__add_new_work_types.sql
sonny-hmcts: yes i confirmed it with emma smith

PR #1412, opened 2025-04-08
Revert "removed duplicated code"

raghera:
Reverts hmcts/wa-task-management-api#1334


raghera:
Not required as was an unrelated issue causing the master failure.

PR #1424, opened 2025-04-28
Rwa 3586

skandala-cgi:
https://tools.hmcts.net/jira/browse/RWA-3586

### Change description ###
RWA-3586 changes

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/wataskmanagementapi/domain/task/Task.java
```diff
@@ -519,5 +522,13 @@ public ZonedDateTime getReconfigureRequestTime() {
     public ZonedDateTime getLastReconfigurationTime() {
         return lastReconfigurationTime;
     }
+
+    public String getTerminationProcess() {
+        return terminationProcess;
+    }
+
+    public void setTerminationProcess(String terminationProcess) {
```
raghera, line 530: Generally we want to avoid setters to keep objects immutable.  So we'd want to remove this when we have switched it on as the only reason it is here is because we need to be able to retrieve the feature flag from LD.

src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/watasks/controllers/PostTaskCompleteByIdControllerTest.java
```diff
@@ -100,6 +101,119 @@ public void should_return_a_204_when_completing_a_task_by_id() {
...
+                taskId,
+                userWithCompletionProcessDisabled.getHeaders()
+            );
+            assertions.taskFieldWasUpdatedInDatabase(
+                taskId, "termination_process", null, userWithCompletionProcessDisabled.getHeaders()
+            );
+            result.then().assertThat()
+                .statusCode(HttpStatus.NO_CONTENT.value());
+            String completionProcess = data[0];
+            String terminationProcess = data[1];
```
raghera, line 194: Do we need this as it does not seem to be used in this test?

raghera (approved):

raghera (approved):

PR #1478, opened 2025-07-21
Task title must default to the Task name during Initiation & Reconfiguration

skandala-cgi:
https://tools.hmcts.net/jira/browse/RWA-4550

### Change description ###
Task title must default to the Task name during Initiation & Reconfiguration

```
[x] Yes
[ ] No
```


src/functionalTest/java/uk/gov/hmcts/reform/wataskmanagementapi/watasks/controllers/PostTaskExecuteReconfigureControllerTest.java
```diff
@@ -1045,6 +1047,7 @@ public void should_reconfigure_camunda_task_attributes_after_validation_reconfig
             .statusCode(HttpStatus.OK.value())
             .and().contentType(MediaType.APPLICATION_JSON_VALUE)
             .and().body("task.id", equalTo(taskId))
+            .body("task.task_title", equalTo("A Task")) //Default task name
```
raghera, line 1050: Is there a scenario where the existing `title` is valid but not equal to the Task `name` and during Reconfiguration you get an empty `title`? In this case the existing `title` should be maintained and not set it to the default value.

Should work like this but not sure if we have an existing test?

raghera (approved):

